PR #12, opened 2020-09-02
Make sure we link to articles in the public handbook

zachmargolis:
- As we migrate we link in to the old handbook. This adds a spec that
  will fail if we continue to link to the old article after already
  migrating it

### DIFF
--- a/spec/support/matchers.rb
+++ b/spec/support/matchers.rb
@@ -24,3 +24,22 @@ RSpec::Matchers.define :have_unique_ids do
   end
 end
 
+RSpec::Matchers.define :link_internally_for_handbook_articles do
+  articles = Set.new
+
+  match do |actual|
+    doc = actual
+
+    doc.css('a[href*="login-handbook.app.cloud.gov"]').each do |a|
+      path = URI(a[:href]).path
+
+      articles << path if file_at(path)
+
+      expect(articles).to be_empty
+    end
+  end
+
+  failure_message do |actual|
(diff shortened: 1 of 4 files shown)


spec/support/matchers.rb
```diff
@@ -24,3 +24,22 @@
...
+      path = URI(a[:href]).path
+
+      articles << path if file_at(path)
+
+      expect(articles).to be_empty
+    end
+  end
+
+  failure_message do |actual|
+    "expected that #{actual.url} would not link to external version of articles in this handbook, but found:\n#{articles.to_a.join("\n")}"
```
aduth, line 43: At this point, would `articles` always be just the single entry that existed at the time we called `expect(articles).to be_empty`? If so, would we want to wait until later to make the assertion if we want to log all of the violations at the same time? Or alternatively, just assign and print the one article we encountered first?

aduth (approved):

spec/support/matchers.rb
zachmargolis: So I think the way RSpec matchers are called, the whole matcher block is "instantiated" for every assertion. The variable is at that outer scope because it needs to be shared between the `match` block and the `failure_message` block, it ends up working fine

zachmargolis: oh wait I see now --- the `expect(articles).to be_empty` needs to move outside that block, d'oh

fixed in 7b555fe38c13df2aa5834c25d173254d5f708723

PR #13, opened 2020-09-02
Bring over Slack article

zachmargolis:
(I removed a few rooms that reference vendors and partner agencies that aren't public yet)


aduth (approved):

ctro (approved):
LGTM

PR #14, opened 2020-09-02
Migrate over Github article

zachmargolis:
- Also bring over PrivateEye from TTS Handbook to mark private
  links as private

Example of private repos:

Screen Shot 2020-09-02 at 4 04 26 PM


_articles/github.md
```diff
@@ -0,0 +1,131 @@
+---
+title: Github
```
aduth, line 2: Nit (existed previously as well):

```suggestion
title: GitHub
```

_articles/github.md
```diff
@@ -0,0 +1,131 @@
...
+### Applications
+
+- [**18f/identity-idp**](https://github.com/18f/identity-idp)<br />
+  [secure.login.gov](https://secure.login.gov)<br />
+  IDP (main application)
+
+- [**18f/identity-pki**](https://github.com/18f/identity-pki)
+  PIV/CAC application
+
+- [**18f/identity-idp-config**](https://github.com/18f/identity-idp-config){:class="private-link"} Service Provider (`service_providers.yml`) and Agency (`agencies.yml`) configuration repo that can be updated and deployed independently of the IDP.
```
aduth, line 21: It's a bit unexpected to me that we'd both manually assign `class="private-eye"`, and then also configure the JavaScript to do so for other links on our behalf. Should we either configure PrivateEye initialization to capture this link pattern? Or conversely, avoid initializing any default URLs to privatize?

aduth (approved):

_articles/github.md
zachmargolis: Hm this is a good point, it's confusing to have both the global list and the specific classes.

I'll unify this by putting it all in the global list, it matches the TTS handbook, and that means if we link in to any of these repos from individual articles, they'll get the lock icons as well

zachmargolis: Updated in 1221d11ecf09aa337aa1a6e0ef9816cb676df8bd

PR #15, opened 2020-09-03
Fix private handbook matcher to allow links to root

zachmargolis:
The repos page links to the private handbook's URL to show what site it goes to, and links to the root.

The "private handbook lint" script makes sure that if we link to the old handbook, the corresponding page should not exist.

However, we kind of need to have a root page, so that's a special case, so I updated the matcher to skip that


mitchellhenke (approved):
👍🏼 👍🏼

PR #19, opened 2020-09-09
Add instructions for how to deploy OIDC, SAML sinatra example apps

zachmargolis:
I wrote down instructions I saved from @solipet in this article


ctro (approved):
LGTM

PR #31, opened 2020-10-07
Onboarding updates

ctro:
Just some random on-boarding updates that I've had queued up


zachmargolis (approved):
LGTM

_articles/onboarding.md
```diff
@@ -51,6 +53,7 @@ Look at this work as a power multiplier, you are helping someone gain a firm fou
   - Sprint ceremonies
   - All-hands / Demo-day
   - Retros / IRLs
+- [Add user to Figma](https://www.figma.com/files/team/893580939040886405/Login.gov/members) if appropriate (e.g. on UX team or implementing interfaces)
```
zachmargolis, line 56: Can we add `figma.com` to the list of private URLs in `_layouts/default.html`? Then it'll get a lil lock icon

ctro: Oh, totally. I didn't realize that's how it worked.  Added.

zachmargolis: I'll make a TODO for myself to extract it to a YAML since that *might* be more discoverable

PR #32, opened 2020-10-07
Add Partnerships specific onboarding tasks

ctro:
This PR should replace Aaron's partnerships-specific onboarding doc (https://docs.google.com/document/d/1ip6oEA7sQjuNUbIrDvO11c1iDxapUyRkXE6SpGdOEak/edit#)

There were only a couple things in your list that weren't already in the public handbook.
When this looks :+1: to you we'll make a deprecation note in your GDoc and add a link to the handbook article: https://handbook.login.gov/articles/onboarding.html

This also begs the question if the linked "partnerships area" in the old, private handbook is still useful.
Aaron, should we move that document over also?

### DIFF
--- a/_articles/onboarding.md
+++ b/_articles/onboarding.md
@@ -60,7 +60,7 @@ Look at this work as a power multiplier, you are helping someone gain a firm fou
 
 ## For Partnerhsips team members
 
-- Check out the [Partnerships Area of the handbook](https://login-handbook.app.cloud.gov/articles/partnerships.html)
+- Check out the [Partnerships Area of the handbook]({{site.baseurl}}/#partnerships)
 - Be sure to join the main Slack channels for our team, `#login` and `#login-partnerships`.
 - Add the following email address to your Google Calendar to see the Login Partnerships Team Calendar: `gsa.gov_6ovul6pcsmgd40o8pqn7qmge5g@group.calendar.google.com`
 - Add the user to [Hubspot](https://app.hubspot.com/settings/5531666/users)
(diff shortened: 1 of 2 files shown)


_articles/onboarding.md
```diff
@@ -58,6 +58,14 @@ Look at this work as a power multiplier, you are helping someone gain a firm fou
 - Familiarize yourself with the [Login.gov Design System](https://design.login.gov/)
 - Familiarize yourself with the [Login.gov UX Drive Folders](https://drive.google.com/drive/folders/12qRTGijG9oOU8FRvZfK30qAN4v8LCzHG)
 
+## For Partnerhsips team members
+
+- Check out the [Partnerships Area of the handbook](https://login-handbook.app.cloud.gov/articles/partnerships.html)
+- Be sure to join the main Slack channels for our team, `#login` and `#login-partnerships`.
+- Add the following email address to your Google Calendar to see the Login Partnerships Team Calendar: `gsa.gov_6ovul6pcsmgd40o8pqn7qmge5g@group.calendar.google.com`
```
zachmargolis, line 65: is this correctly permissioned that it's OK to post in the public handbook?

_articles/onboarding.md
```diff
@@ -58,6 +58,14 @@ Look at this work as a power multiplier, you are helping someone gain a firm fou
 - Familiarize yourself with the [Login.gov Design System](https://design.login.gov/)
 - Familiarize yourself with the [Login.gov UX Drive Folders](https://drive.google.com/drive/folders/12qRTGijG9oOU8FRvZfK30qAN4v8LCzHG)
 
+## For Partnerhsips team members
+
+- Check out the [Partnerships Area of the handbook](https://login-handbook.app.cloud.gov/articles/partnerships.html)
```
zachmargolis, line 63: this article is also kind of partnerships specific, maybe we should recategorize it from appdev to parterships? https://handbook.login.gov/articles/deploying-sp-to-prod.html

zachmargolis (approved):
LGTM

_articles/onboarding.md
ctro: 👍  makes sense to me, changed.

_articles/onboarding.md
ctro: I believe so.  "Access Permissions" are set to "Make available for GSA.gov", not "Make available to public"

PR #33, opened 2020-10-08
Move review document to new handbook

ctro:
Here is the PR removing from the private handbook: https://github.com/18F/identity-handbook-private/pull/164


zachmargolis (approved):
LGTM

PR #37, opened 2020-10-28
Fix list formatting for engineering definition of done

aduth:
"Database indexes exist for any new queries" was not formatted as a list item, and would appear as combined with the previous list entry

Before|After
---|---
before|Screen Shot 2020-10-28 at 9 17 52 AM


pauldoomgov (approved):
LGTM

PR #39, opened 2020-10-29
Update login-product Slack channel references to login-appdev

aduth:
**Why**: Channel was renamed


zachmargolis (approved):
LGTM! Can we update in the `slack.md` article too?

zachmargolis:
> Can we update in the `slack.md` article too?

Whoops just realized that the channel is not listed in that article at all 🙈. So no need to change. But bonus....add it?

aduth:
>So no need to change. But bonus....add it?

Bonus points ⭐  43874c6

More bonus: Channel typos / incorrect links (a0c34e1, 6e9b8e6, d0516c6)

PR #47, opened 2020-12-02
Onboarding updates

ctro:
- Added links to authentication and identity proofing intro videos
- Added more Slack groups and onboarding reminders

### DIFF
--- a/_articles/slack.md
+++ b/_articles/slack.md
@@ -28,7 +28,10 @@ These handles ping oncall engineers, use for emergencies or urgent items.
 - `@login-partnerships-team` Group to address login.gov partnerships team members
 - `@login-product-team` Group to address login.gov product managers
 - `@login-ux-team` Group for the design/UX team
+- `@login-appdev-team` Group for application development teams
 - `@login_leads` login team leads
+- `@login-feds` Group for login.gov federal FTE employees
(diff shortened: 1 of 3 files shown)


_articles/slack.md
```diff
@@ -28,7 +28,10 @@ These handles ping oncall engineers, use for emergencies or urgent items.
 - `@login-partnerships-team` Group to address login.gov partnerships team members
 - `@login-product-team` Group to address login.gov product managers
 - `@login-ux-team` Group for the design/UX team
+- `@login-appdev-team` Group for application development teams
```
zachmargolis, line 31: should we remove `@login-eng` since that is also the "app team" ?

zachmargolis (approved):
LGTM, one quick note, not a blocker

orenyk (approved):
Agree with @zachmargolis, looks good (and I made sure that the video links were correct). :shipit:

_articles/slack.md
ctro: `login-eng` looks like just engineers.  There was request for an `appdev-team` group that includes all members of Ada, Grace, and Katherine. (UX, Product, etc is included).    How can I make that more clear?

zachmargolis: Oh I have no idea, maybe let's leave in both for the time being then

PR #50, opened 2020-12-07
Guidance on disabling encrypted responses locally

solipet:


zachmargolis (approved):
LGTM, is there a param (or SP config) we pass in to the SAML IDP gem we could disable instead of having to muck with the gem?

PR #51, opened 2020-12-09
List accounts that we use, link to appendix page for access (LG-3817)

zachmargolis:
Please take a look at the appendix page too!

### DIFF
--- a/_articles/offboarding.md
+++ b/_articles/offboarding.md
@@ -25,13 +25,14 @@ Review the [Leaving TTS page in the TTS Handbook](https://handbook.tts.gsa.gov/l
 - [Using the JIRA Portal](https://cm-jira.usa.gov/servicedesk/customer/portal/11), choose `Application Access` and request that the user be removed from the Login.gov project (and deactivated if they are no longer working for GSA).
 - Use the [TTS Slack Form](https://goo.gl/forms/mKATdB9QuNo7AXVY2) to submit user modification
 - Remove from [login.gov Slack groups]({{site.url}}/articles/slack.html).
-- [Remove user from Login.gov Google Groups](https://groups.google.com/a/gsa.gov/forum/#!myforums)
-- [Remove the user from Hubspot](https://app.hubspot.com/settings/5531666/users)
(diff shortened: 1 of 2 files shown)


_articles/offboarding.md
```diff
@@ -25,13 +25,14 @@ Review the [Leaving TTS page in the TTS Handbook](https://handbook.tts.gsa.gov/l
...
 - Use the [TTS Slack Form](https://goo.gl/forms/mKATdB9QuNo7AXVY2) to submit user modification
 - Remove from [login.gov Slack groups]({{site.url}}/articles/slack.html).
-- [Remove user from Login.gov Google Groups](https://groups.google.com/a/gsa.gov/forum/#!myforums)
-- [Remove the user from Hubspot](https://app.hubspot.com/settings/5531666/users)
-- [Remove user from Figma](https://www.figma.com/files/team/893580939040886405/Login.gov/members)
-- Remove the user from the [`gsa-login-prototyping` cloud.gov org](https://dashboard.fr.cloud.gov/cloud-foundry/2oBn9LBurIXUNpfmtZCQTCHnxUM/organizations/fc240d49-f678-4325-8384-c88d92d60982/users)
-- Remove the user as admins from any dashboards
-  - [login.gov dashboard](https://dashboard.int.identitysandbox.gov)
-  - [search.gov dashboard](https://search.gov)
+- Remove from [all accounts]({{site.url}}/articles/accounts.html)
```
aduth, line 28: Not something to address here, but in the course of reviewing the Federalist preview, I observe that using absolute URLs makes it difficult, since these point to the live site. Relative URLs (or Jekyll `baseurl`) would be easier.

aduth (approved):
👍

_articles/offboarding.md
mitchellhenke: 👍🏼 

I think there is something like a `link` helper that also validates the page exists

zachmargolis: Ooo I'll follow up with a PR to use the `link` helper shortly

PR #55, opened 2020-12-10
Update secops-incident-response-guide.md

mzia:
remove Situation Lead to what we are all familiar with Incident Commander. Silly we had to change it for GSA IR team...


zachmargolis (approved):
Love the name

PR #56, opened 2020-12-11
Note on skipping encryption

solipet:
Replaced instructions on how to hack around the encrypted response with a note about the new option to skip encryption.


zachmargolis (approved):
LGTM

PR #57, opened 2020-12-14
Add note about default PA environment being prod... and more!

pauldoomgov:
* AC for Team Radia issues should be checked when release to `prod` is complete
* Also adds links to IR Checklist and Troubleshooting Quick Reference for infra on-call


zachmargolis (approved):
LGTM

PR #59, opened 2020-12-15
guidance on how to fix Invalid Signature

solipet:
Somehow the `dev` env was configured with the same cert signature as in `int`, so we were getting an Invalid Signature error in the SAML Sinatra sample app in `dev`. This walks through what I had to do to resolve it.


zachmargolis (approved):
more SAML lore LGTM

PR #60, opened 2020-12-21
Remove ASCII async diagram, use pretty image one

zachmargolis:
I had to make this prettier diagram for the SSP, so seemed easy enough to include it here


mzia (approved):
Thank you! LGTM.

PR #62, opened 2020-12-21
Pull cost recovery into a separate article

orenyk:
Also fix a few typos

### DIFF
--- a/_articles/cost-recoverability.md
+++ b/_articles/cost-recoverability.md
@@ -1,0 +1,23 @@
+---
+title: Cost Recoverability
+layout: article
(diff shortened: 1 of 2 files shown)


_articles/cost-recovery.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Cost Recoverability
```
zachmargolis, line 2: let's make the title and URL match? Recovery vs Recoverability?

zachmargolis (commented):
LGTM

zachmargolis (approved):

PR #63, opened 2020-12-21
Add openssl command line recipes article (LG-3931)

zachmargolis:
Because the OpenSSL CLI is so hard to remember that every team I've been on has needs some version of one of these.

cc @jmhooper there are the ones I could remember you using when we debugged the cert issue the other day, but if you have more I'm happy to add them!


pauldoomgov (approved):
YES!

PR #64, opened 2020-12-28
Add high-level Partnerships RACI data

ctro:
This is a high-level summary of the data that the partnerships team put together here: https://docs.google.com/spreadsheets/d/1GV9FOdXtbY4cmMupPMMz0nrhTgTWMVZkzWPXO18bEZw/edit#gid=0

This page is pretty long, and consists of multiple tables. Putting everything into one table does not work well horizontally.
Does this seem OK, or is there a better way?

### DIFF
--- a/_articles/partnerships-raci.md
+++ b/_articles/partnerships-raci.md
@@ -13,149 +13,171 @@ It is a [responsibility assignment matrix](https://en.wikipedia.org/wiki/Respons
 
 Documenting our matrix is an on-going process. The partnerships team continually updates the [detailed source of truth in this Google Doc](https://docs.google.com/spreadsheets/d/1GV9FOdXtbY4cmMupPMMz0nrhTgTWMVZkzWPXO18bEZw/edit#gid=0). When new high-level trends are identified they are added to the high-level overview table below.
 
-## High level overview
+## Agreements - Template Change
 
-1. [Agreements - Template Change](#agreements-template)
-2. [Agreements - Establishing/Adjusting Terms](#agreements-terms)
-3. [Financial](#financial)
-4. [Managing the Service Providers in Prod](#service-providers)
-5. [Product](#product)
-6. [Shipping Public/Client-Facing Materials](#client-materials)
-7. [Technical](#technical)
-8. [Service](#service)
+### Responsible
 
-<a name="agreements-template"></a>
+1. Login Partnerships
+2. TTS Agreements
+3. Partner IAA POC
 
-| Agreements - Template Change |  |
-|---|-----|
-| **Login** Leadership | A |
-| **Login** Finance | C |
-| **Login** Partnerships | R |
-| **Login** Product | I |
-| **Login** Infrastructure | I |
-| **Login** UX |  |
-| **TTS** Leadership | I |
-| **TTS** OGC | C |  
-| **TTS** Agreements | R |  
-| **Partner** IAA POC | R  |
-| **Partner** App POC |  |
-| **Partner** OGC | C |
+### Accountable
 
-<a name="agreements-terms"></a>
+1. Login Leadership
 
-| Agreements - Establishing/Adjusting Terms |  |
-|---|-----|
-| **Login** Leadership | I |
-| **Login** Finance | C |
-| **Login** Partnerships | R/A |
-| **Login** Product |  |
-| **Login** Infrastructure |  |
-| **Login** UX |  |
-| **TTS** Leadership |  |
-| **TTS** OGC | C |  
-| **TTS** Agreements | R/A |  
-| **Partner** IAA POC | R/A  |
-| **Partner** App POC |  |
-| **Partner** OGC |  |
+### Consulted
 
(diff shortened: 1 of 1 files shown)


ctro:
@colinmurphy01 tagging you here cause you might be interested

_articles/partnerships-raci.md
```diff
@@ -0,0 +1,161 @@
...
+3. [Financial](#financial)
+4. [Managing the Service Providers in Prod](#service-providers)
+5. [Product](#product)
+6. [Shipping Public/Client-Facing Materials](#client-materials)
+7. [Technical](#technical)
+8. [Service](#service)
+
+<a name="agreements-template"></a>
+
+| Agreements - Template Change |  |
```
orenyk, line 29: Is there any way to specify fixed/consistent widths for the columns of all of the tables? The changing widths made it a bit hard to parse. Additionally, I wonder if there's a better way to present this - maybe instead of just inverting the RACI table we instead just have four columns for each task/activity and list the groups that fall under each category? In other words:

---

|   | Responsible | Accountable | Consulted | Informed |
|---|---|---|---|---|
| Agreements - Template Change | 1. **Login** Partnerships<br />2. **TTS** Agreements<br />3. **Partner** IAA POC | 1. **Login** Leadership | 1. **Login** Finance<br />2. **TTS** OGC | 1. **Login** Product<br />2. **Login** Infrastructure<br />3. **TTS** Leadership |

---

The formatting is still a bit wonky but that might make it a bit easier to grok? At that point you might also be able to dispense with tables and instead do something like

---

### Agreements - Template change

This is a description of what a template change to one of our agreements is. Which agreements are included? What constitutes a change (i.e. everything)?

#### Responsible

1. **Login** Partnerships
2. **TTS** Agreements
3. **Partner** IAA POC

#### Accountable

1. **Login** Leadership

#### Consulted

1. **Login** Finance
2. **TTS** OGC

#### Informed

1. **Login** Product
2. **Login** Infrastructure
3. **TTS** Leadership

---

Curious to hear what you think!

orenyk (commented):
Had some ideas for presenting the data a little more cleanly, let me know what you think!

_articles/partnerships-raci.md
ctro: Hmm, yeah.  I'm not sure 🤔 .  I guess it depends on how you all are going to use this.   I was thinking it would be useful to easily compare the differences between topics -- which seems easier in table-land.   That might be the wrong way to think about it though.   I'm curious to hear what the rest of the partnerships team thinks.

I'm super happy to change this to be what y'all want.  I agree that with your examples, a bullet list is more readable than a table.

Also, I'm not sure if there is a way to specify equal width tables.  Might be able to write some raw HTML, or add some CSS, or maybe even, gasp, nest the tables in one big table?

orenyk: Yeah my guess would be to wrap it in a div with a specific class and include more specific widths for the overall table and the columns.

Re: comparison as the goal, I think once the table is split up it's hard to draw the comparison regardless since you're scrolling so much, but maybe I'm wrong. I also wonder if the more common use case will be to look up "for this thing I'm trying to do, who do I work with / contact / etc?", but I'd definitely want the rest of the team to weigh in as well 😄.

ctro: @orenyk I switched the content over to lists from tables, is this about how you envisioned?

_articles/partnerships-raci.md
```diff
@@ -0,0 +1,183 @@
...
+## Partnerships RACI framework
+
+RACI stands for **R**esponsible, **A**ccountable, **C**onsulted, **I**nformed.
+It is a [responsibility assignment matrix](https://en.wikipedia.org/wiki/Responsibility_assignment_matrix) framework for understanding various roles in completing tasks or deliverables for a project.
+
+## Full documentation
+
+Documenting our matrix is an on-going process. The partnerships team continually updates the [detailed source of truth in this Google Doc](https://docs.google.com/spreadsheets/d/1GV9FOdXtbY4cmMupPMMz0nrhTgTWMVZkzWPXO18bEZw/edit#gid=0). When new high-level trends are identified they are added to the high-level overview table below.
+
+## Agreements - Template Change
```
zachmargolis, line 16: The sidenav is really noisy with these headings, I have two ideas:

1. set `toc_h_max: 2` which means the table of contents will only go down to h2s, so the "responsible" "accountable" "consulted" etc don't clutter the sidenav
2. Or just turn the Responsible, Accountable, etc headers into just plain bolds, `**Responsible**`?

Example before:
Screen Shot 2021-01-20 at 10 30 18 AM

Example after:
Screen Shot 2021-01-20 at 10 31 08 AM

zachmargolis (approved):
LGTM, some nav suggestions

_articles/partnerships-raci.md
ctro: Ooh, nice I didn't know about toc_h_max, I added that in.

PR #72, opened 2021-01-13
We no longer manually deploy to Staging

solipet:
Removed the instructions to deploy to staging. Staging is updated daily to the `master` branch (both idp and pki)


zachmargolis (approved):
LGTM

PR #74, opened 2021-01-19
Oncall update: PKI Certs

zachmargolis:
Also cleaned up formatting for the the severities section while I'm here


ctro (approved):
:shipit: LGTM thanks Zach.

PR #76, opened 2021-01-20
Add collected onboarding updates

ctro:
- Notes for VPN and Admin access
- Instructions for feds-only and contractors-only Google Groups
- Add security overview from Summit.


zachmargolis (approved):
LGTM

PR #82, opened 2021-02-01
Migrate SAML cert rotation article

zachmargolis:
Copied from the private handbook


orenyk (approved):
Looks good, didn't do an in-depth read of the content but seems good (and verified on the preview site). :shipit:

PR #84, opened 2021-02-04
Add new-branch flag for rollback instructions

aduth:
**Why**: If one is to assume the set of commands are for a revert branch yet to be created, omitting the flag would result in an error "error: pathspec 'revert-rc-123' did not match any file(s) known to git".


zachmargolis (approved):
LGTM

PR #95, opened 2021-03-04
Add note on waiting for devops deploy before updating SAML config

pauldoomgov:
Clarifies the blocker for deployment of updated Chef code before the new SAML certificate is actually ready for use.


zachmargolis (approved):
LGTM

solipet (approved):

PR #100, opened 2021-03-11
Remove page "Deploying the Sample Apps"

aduth:
**Why**: Because they deploy automatically on merge to main via CircleCI. Neat!

See:

- https://github.com/18F/identity-oidc-sinatra/pull/90
- https://github.com/18F/identity-saml-sinatra/pull/77


zachmargolis:
instead of deleting, what if the page said that they deploy automatically and linked to their circleCI configs?

zachmargolis (approved):

aduth:
> instead of deleting, what if the page said that they deploy automatically and linked to their circleCI configs?

Works for me 👍  325cfb0

PR #101, opened 2021-03-12
Add OpenSSL recipes for converting pfx (pkcs12) to PEM

zachmargolis:
I hope that in 3 years these just work so nobody has to blindly futz with OpenSSL


aduth (approved):
👍

mitchellhenke (approved):

PR #111, opened 2021-04-22
Deprecate lambda jobs article

zachmargolis:
I don't know what our approach is for old articles, I figured this one was kind of useful as a reference so I opted to deprecated it rather than direct delete? I know we linked to it from some previous SCR docs

So I created a new concept of deprecating an article which adds:

| item | screenshot |
| --- | --- |
| grayed out link and deprecation note on homepage | Screen Shot 2021-04-22 at 9 27 43 AM |
| deprecated title on article page | Screen Shot 2021-04-22 at 9 27 35 AM |

### DIFF
--- a/_articles/appdev-lambda-jobs.md
+++ b/_articles/appdev-lambda-jobs.md
@@ -3,7 +3,18 @@ title: "Background Jobs: Lambda"
 description: "Overview of and launch checklist for our async Lambda workers"
 layout: article
 category: "AppDev"
+deprecated: true
 ---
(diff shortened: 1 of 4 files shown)


jmhooper (approved):
Nice, I think that works. Good call on deprecating it instead of deleting it.

We may want to consider filtering them and moving them to a separate category at some point. Maybe if we start having too many. Right now we only have 1 deprecated article though so I think we can wait and see if we actually get to the point.

PR #115, opened 2021-05-03
Update configuration to use NPM instead of Yarn

aduth:
**Why**: Jekyll and CircleCI configurations assume a Yarn lockfile which doesn't exist. It (as well as a package-lock.json) had existed prior to #2. There should be only one. Since there's already an NPM lockfile, use it for caching, and exclude from Jekyll build.


zachmargolis (approved):
LGTM

PR #116, opened 2021-05-03
Rename build workflow to omit "deploy" mention

aduth:
**Why**: There is no deployment as part of CircleCI. This happens in Federalist. This was a remnant from the old private version of the handbook, which included a deployment step.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,6 @@ jobs:
             - "_site/*"
 workflows:
   version: 2
-  # run on pull requests and after merges
-  build_and_deploy:
+  build:
     jobs:
       - build


.circleci/config.yml
```diff
@@ -49,7 +49,6 @@ jobs:
             - "_site/*"
 workflows:
   version: 2
-  # run on pull requests and after merges
-  build_and_deploy:
+  build:
```
zachmargolis, line 52: this is gonna have to be updated in GH branch protections

zachmargolis (approved):
LGTM

.circleci/config.yml
aduth: Thanks 👍

PR #118, opened 2021-05-05
Update table of contents to indent less

zachmargolis:
The indentation was causing a lot of weird line breaks, I figure we can do more with less

| before | after |
| --- | --- |
|  Screen Shot 2021-05-05 at 10 17 38 AM | Screen Shot 2021-05-05 at 10 17 45 AM |


assets/css/main.scss
```diff
@@ -37,4 +37,8 @@
 
 .deprecated-link {
   color: gray;
+}
+
+.table-of-contents ul {
+  padding-inline-start: 15px;
```
aduth, line 43: Not sure if there is a strong reason to use `padding-inline-start` over `padding-left` , which is also [supported in Internet Explorer](https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline-start#browser_compatibility)?

```suggestion
  padding-left: 15px;
```

aduth (approved):
👍 

Ideally we should try to move to the [Side navigation component](https://design.login.gov/components/side-navigation/).

assets/css/main.scss
zachmargolis: I picked that because it's what showed up in the Chrome inspector, `padding-left` is the one I'm more familiar with, will go with that

zachmargolis:
> Ideally we should try to move to the [Side navigation component](https://design.login.gov/components/side-navigation/).

Good call! I updated the TOC generator, which now lets us annotate sublists, so we can use USWDS. It doesn't seem to have `current` support, but is close enough!

Screen Shot 2021-05-05 at 11 17 35 AM

aduth (approved):
Nice! 🎉 

Some stylistic issues related to using an older version of the design system, but that's tracked in LG-3759.

PR #129, opened 2021-06-14
Remove reference to slim files

mitchellhenke:
They don't exist any longer 🙂


zachmargolis (approved):
Slim? never heard it it

PR #133, opened 2021-06-22
Workers: Clarify presigned URL expiration behavior

aduth:
**Why**: So that the expiration behavior is clearly defined.

See Slack thread: https://gsa-tts.slack.com/archives/C01D7MWNE64/p1623333374015500


zachmargolis (approved):
LGTM

PR #138, opened 2021-07-08
Add worker host type to deploy guide

zachmargolis:
New worker type means new command to copy-paste


aduth (approved):
Are workers enabled in production? 🤔  Do they need to be for this guidance to make sense?

mitchellhenke (approved):

mitchellhenke:
@aduth they're in production, but not enabled

PR #140, opened 2021-07-08
Add a glossary page

zachmargolis:
Earlier today, I was like "what the heck is an [iaa_gtc](https://github.com/18F/identity-idp/blob/main/app/models/agreements/iaa_gtc.rb)??" so now I'm making an article so somebody else could maybe look up the same thing in the future


orenyk (approved):
Great idea!

PR #152, opened 2021-07-27
Add environments stub page

aduth:
**Why**: So that we can have a reference for which environments exist and the differences between them.

Redirects to an internal Google Doc.

### DIFF
--- a/_articles/sandboxes-and-staging-environments.md
+++ b/_articles/sandboxes-and-staging-environments.md
@@ -1,0 +1,7 @@
+---
+title: Environment Descriptions
+description: Listing of environments and the differences between them
+layout: article
+category: AppDev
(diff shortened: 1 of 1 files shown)


_articles/sandboxes-and-staging-environments.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Environment Descriptions
+description: Listing of environments and the differences between them
```
zachmargolis, line 3: for "SEO" can we add some extra bits?

```suggestion
description: Listing of environments and the differences between them, like prod, pt, dm, int or dev
```

(this is both for cmd-F searching as well as the search box up top)

zachmargolis (approved):
LGTM

_articles/sandboxes-and-staging-environments.md
aduth: Good idea! This definitely aligns to how I imagine I'd attempt to recall this information in the future.

PR #159, opened 2021-09-03
Fix typo in aws-vault command

aduth:
**Why**: Because `production-power` isn't a valid profile name.

>aws-vault: error: exec: Error getting temporary credentials: profile production-power: credentials missing


zachmargolis (approved):
LGTM 🤦

PR #160, opened 2021-09-10
Update reporting dashboard image caption

aduth:
**Why**: Because it's inaccurate, and likely a copy-paste error from similar [Background Jobs article / image](https://handbook.login.gov/articles/appdev-proofing-ruby-worker-jobs.html).

See also: https://github.com/18F/identity-handbook/pull/158#discussion_r697110515


zachmargolis (approved):
LGTM

PR #163, opened 2021-09-22
Add design system link styling

aduth:
**Why**: So that links are styled using the design system, instead of browser defaults.

An alternative can be to configure [the `$theme-global-link-styles` SASS variable](https://github.com/uswds/uswds/blob/019eda3109a9074923e60326a4aadb4c6b56d04a/src/stylesheets/global/_typography.scss#L17-L19). However, this applies to all links in the layout, which may have unexpected repercussions. Instead, extending Kramdown ensures that styles are only applied to parsed Markdown content.

Before|After
---|---
image|image

Kramdown monkey-patching reference: https://github.com/gettalong/kramdown/blob/0b0a9e0/lib/kramdown/parser/kramdown/link.rb#L37-L51

### DIFF
--- a/_plugins/content_typography.rb
+++ b/_plugins/content_typography.rb
@@ -1,0 +1,12 @@
+module Kramdown
+  module Parser
+    class Kramdown
+      prepend(Module.new do
+        def add_link(el, *args)
+          el.attr['class'] = [*el.attr['class'], 'usa-link'].join(' ') if el.type == :a
+          super(el, *args)
+        end
+      end)
+    end
+  end
+end


zachmargolis (approved):
LGTM

_plugins/content_typography.rb
```diff
@@ -0,0 +1,12 @@
+module Kramdown
+  module Parser
+    class Kramdown
+      prepend(Module.new do
+        def add_link(el, *args)
+          el.attr['class'] = [*el.attr['class'], 'usa-link'].join(' ') if el.type == :a
+          super(el, *args)
```
aduth, line 7: Could serve as motivation to bump to Ruby >=2.7 😄 

```suggestion
        def add_link(el, ...)
          el.attr['class'] = [*el.attr['class'], 'usa-link'].join(' ') if el.type == :a
          super(el, ...)
```

PR #169, opened 2021-10-01
Add leave guidance

ctro:
Adding basic leave guidance

### DIFF
--- a/_articles/leave-guidance.md
+++ b/_articles/leave-guidance.md
@@ -1,0 +1,15 @@
+---
+title: Leave Guidance
+layout: article
+category: Team
+---
+
+## Login.gov Leave Guidance
+
+Please follow [leave guidance in the TTS Handbook](https://handbook.tts.gsa.gov/travel-and-leave/leave/)
+
+Additionally, please do the following
+
+- Give at least one-sprint, but preferably two-sprint notice if you are going to be out for more than a couple days
+- Put your OOO time in the Shared Login.gov Calendar and decline meetings
+- List your back-ups in GMail auto-replies and Slack profile


sharms (approved):
LGTM!

pauldoomgov (approved):
LGTM

_articles/leave-guidance.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Login.gov Leave Guidance
+
+Please follow [leave guidance in the TTS Handbook](https://handbook.tts.gsa.gov/travel-and-leave/leave/)
+
+Additionally, please do the following
+
+- Give at least one-sprint, but preferably two-sprint notice if you are going to be out for more than a couple days
+- Put your OOO time in the Shared Login.gov Calendar and decline meetings
+- List your back-ups in GMail auto-replies and Slack profile
```
aduth, line 15: A question from ignorance: Who are my back-ups? Does this apply to me?

ctro: If you don't know who they are, then you probably don't have any.  
If there's something you're responsible for and you'll be away for a week, consider talking to someone about being your back-up.   Maybe "Any questions about  LGDesignSystem should be directed to Nick", etc.

How could we write that to be a little more clear?

aduth: That makes a lot of sense to me just as you've written there. Would it be too lengthy to include in-line? Maybe as a sub-bullet point?

```suggestion
- List your back-ups in GMail auto-replies and Slack profile. If you don't know who they are, then you probably don't have any. If there's something you're responsible for and you'll be away for a week, consider talking to someone about being your back-up.
```

PR #171, opened 2021-10-01
Update leave page

ctro:
Be more specific about back-ups


aduth (approved):
Thanks, much clearer 👍

PR #173, opened 2021-10-13
Document GitLab/Login.gov auth integration

akrito:
(and fix a tiny typo)

### DIFF
--- a/_articles/gitlab.md
+++ b/_articles/gitlab.md
@@ -18,7 +18,15 @@ openssl req -nodes -x509 -days 365 -newkey rsa:2048 -keyout private.pem -out pub
 ```
 curl -s https://idp.int.identitysandbox.gov/api/saml/metadata2021 \
 | xml sel -N x="http://www.w3.org/2000/09/xmldsig#" -t -v '(//x:X509Certificate)[1]' \
-| openssl x509 -in ~/idp_staging.crt -noout -fingerprint | sed -E 's/.*=//'
+| sed '1i\
+-----BEGIN CERTIFICATE-----
+' \
+| sed '$a\
+-----END CERTIFICATE-----
+' \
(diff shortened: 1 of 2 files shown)


_articles/gitlab.md
```diff
@@ -0,0 +1,35 @@
...
+description: "GitLab Setup"
+layout: article
+category: Infrastructure
+---
+
+## Authentication
+
+GitLab leverages OmniAuth to allow users to sign in using a variety of sersvices, including Login.gov (via SAML). To configure this:
+
+1. Generate a cert and private key by following the instructions at https://developers.login.gov/testing/#creating-a-public-certificate:
```
zachmargolis, line 12:
```suggestion
1. Generate a cert and private key by following the instructions at <https://developers.login.gov/testing/#creating-a-public-certificate>:
```

Add `<>` so it turns into linky text

_articles/gitlab.md
```diff
@@ -0,0 +1,35 @@
...
+```
+openssl req -nodes -x509 -days 365 -newkey rsa:2048 -keyout private.pem -out public.crt
+```
+
+1. Copy the IDP sandbox signing certficate from https://developers.login.gov/saml/ and get its fingerprint:
```
zachmargolis, line 17: 1. linky text
2. typo (certficate/certificate)
3. how do we get from that URL to `idp_sanbox.crt`? Do we save something on the page as that?

```suggestion
1. Copy the IDP sandbox signing certificate from <https://developers.login.gov/saml/> and get its fingerprint:
```
openssl x509 -in idp_sandbox.crt -noout -fingerprint | sed -E 's/.*=//'
```

akrito: Had to massage the output a bit to look exactly how openssl expects it

PR #175, opened 2021-10-21
Add some notes on contacting fellow on-call when things get rough

pauldoomgov:
A suggestion from our Let's Encrypt cert expiration incident review


zachmargolis (approved):
❤️ teamwork makes the dream work

mitchellhenke (approved):

PR #180, opened 2021-11-04
Remove login-eng from Slack team groups

aduth:
**Why**: It appears to not exist anymore. login-appdev-team is the user group for application development team, and is already mentioned further down in the list.


zachmargolis (approved):
LGTM

PR #181, opened 2021-11-08
Update Team Katherine's description

nickttng:
Updating that Team Katherine now works on the IdP too.

Background context: I will be onboarding a new TK member, so I thought that this should be updated.


zachmargolis (approved):
LGTM

PR #183, opened 2021-11-09
Add query for IAL2 users

orenyk:


stevegsa (approved):

_articles/queries.md
```diff
@@ -14,3 +14,11 @@ Returns the number of fully registered users.
 ```ruby
 RegistrationLog.where.not(registered_at: nil).count
 ```
+
+## IAL2 Users
+
+Returns the number of users with IAL2 credentials.
+
+```ruby
+Profile.where(active:true).count
```
zachmargolis, line 23:
```suggestion
Profile.where(active: true).count
```
Small spacing nit

zachmargolis (approved):
LGTM! Thanks for adding

_articles/queries.md
orenyk: Good catch, thanks!

PR #189, opened 2021-11-19
LG-5185: Add SMS and Voice to vendor outage response process

aduth:
**Why:** Since it is now possible to configure messaging to be shown in these flows in the case of an SMS or voice outage (Pinpoint).


zachmargolis (approved):
LGTM

PR #190, opened 2021-11-24
Add command to extend Postgres timeout

orenyk:


_articles/queries.md
```diff
@@ -5,6 +5,14 @@ layout: article
 category: "AppDev"
 ---
 
+## Query timeout
+
+For queries that take longer than the default timeout (2.5s), run the following command to extend the timeout.
```
zachmargolis, line 10: maybe add a note like "please double check that you're on a read replica before running potentially expensive queries"

zachmargolis (approved):
LGTM!

_articles/queries.md
orenyk: Done!

PR #191, opened 2021-12-15
Update user queries

zachmargolis:
- Adds a new "total accounts created" (slightly different from "fully registered IAL 1")
- Adds forms of queries that can be used to approximate historical values (account deletes may mess with all of those, which is why they are not exact)


orenyk (approved):
Great improvement!

PR #196, opened 2021-12-29
Add an article with the daily deploy schedule

zachmargolis:
- And a table of what time it really is locally because time zones
  are really hard

### DIFF
--- a/_articles/daily-deploy-schedule.md
+++ b/_articles/daily-deploy-schedule.md
@@ -17,7 +17,7 @@ These apps are deployed on weekdays:
 
 In these lower environments
 
-- dev
+- dm
 - int
 - staging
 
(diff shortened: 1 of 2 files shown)


_articles/daily-deploy-schedule.md
```diff
@@ -0,0 +1,31 @@
...
+
+- IDP
+- PKI
+- Dashboard
+
+### Environments
+
+In these lower environments
+
+- dev
```
mitchellhenke, line 20:
```suggestion
- dm
```

mitchellhenke (approved):
👍🏼 

Just one minor correction

_articles/daily-deploy-schedule.md
zachmargolis: are you saying replace `dev` with `dm` or just add `dm`?

mitchellhenke: replace `dev` with `dm`

`dev` is on the more frequent recycle schedule

zachmargolis: oh I didn't even know we had a more frequent schedule 😓

PR #207, opened 2022-01-26
Update acceptance thread page

zachmargolis:
- New clearer diagram from Nick, link to Figma
- Updated text, added description


aduth (approved):
Nice 👍

PR #209, opened 2022-01-28
Update dependencies

zachmargolis:
- Run `npm audit fix`
- Update identity-style-guide, this includes a font change

I spot checked a bunch of pages, things seemed to render fine, I was worried about the alert component but that seems to be fine as well

| before | after |
| ---- | ---- |
| Screen Shot 2022-01-27 at 4 55 18 PM | Screen Shot 2022-01-27 at 4 55 10 PM |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,44 @@
 {
+  "name": "identity-handbook",
+  "lockfileVersion": 2,
   "requires": true,
-  "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


aduth (approved):
Looks good 👍  I went through [the breaking changes](https://github.com/18F/identity-style-guide/blob/main/CHANGELOG.md) and didn't spot anything I think should be an issue here.

PR #213, opened 2022-02-03
add How to Manage AWS Lambda Functions

mariuszmspace:


mariuszmspace:
Hi, I just thought it would be great to have a resource for lambda on this handbook as it is for other things in that file:)

PR #225, opened 2022-03-15
Analytics Events Feedback, Lints

zachmargolis:
Adds feedback from #218, adds lints

I was on the fence about adding some sort of tests for this (unit like we haven identity-reporting?) end-to-end like we have in identity-site?), for the time being opted not to.


package.json
```diff
@@ -5,5 +5,15 @@
     "htm": "^3.1.0",
     "identity-style-guide": "^6.3.0",
     "preact": "^10.6.6"
+  },
+  "devDependencies": {
+    "@18f/eslint-plugin-identity": "^2.0.0",
+    "eslint": "^8.11.0",
+    "eslint-plugin-import": "^2.25.4",
+    "eslint-plugin-prettier": "^4.0.0",
+    "prettier": "^2.5.1"
```
aduth, line 14: I dunno if certain editor integrations rely on the file presence as a trigger, but maybe it could be good to include a `.prettierrc` configuration, even if empty (`{}`) ?

assets/js/analytics-events.js
```diff
@@ -42,68 +45,70 @@ const urlify = new function() {
...
+    aria-label="Anchor"
+    data-anchorjs-icon=${icon}
+    id=${slug}
+    href=${`#${slug}`}
+    style="font: 1em / 1 anchorjs-icons; padding-left: 0.375em;"
+  ></a>`;
 }
 
-function Example({ event_name, attributes }) {
+function Example({ event_name: eventName, attributes }) {
```
aduth, line 73: Minor: It would be nice if property names were normalized to JS conventions immediately following API response, so that the component can expect to receive props as named accordingly.

aduth (approved):
Didn't test, but LGTM from what I see 👍

assets/js/analytics-events.js
```diff
@@ -175,35 +182,45 @@ function ErrorPage({ error, url }) {
...
-    render(html`<${Events} events=${events} />`, container);
-    render(html`<${Sidenav} events=${events} />`, sidenav);
-
-    const headerToReplace = document.querySelector('#event-list');
-    if (headerToReplace) {
-      headerToReplace.hidden = true;
-    }
-  })
-  .catch((error) => render(html`<${ErrorPage} url=${eventsUrl} error=${error} />`, container));
+export function LoadAnalyticsEvents() {
```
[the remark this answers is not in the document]
aduth: Just to confirm: This will also need to be updated at call site.

assets/js/analytics-events.js
aduth: Reference: https://airbnb.io/javascript/react/#props

package.json
zachmargolis: added: 53e5ff5294634caafda5283b54fb0cd0b81861a1

PR #226, opened 2022-03-15
Add previous event names to analytics event documentation

zachmargolis:
Related: https://github.com/18F/identity-idp/pull/6063
Screen Shot 2022-03-15 at 8 53 29 AM

### DIFF
--- a/assets/js/analytics-events.js
+++ b/assets/js/analytics-events.js
@@ -111,7 +111,7 @@ function Example({
     .join("\n      ")
     .replace(/ {2}$/, ""); // fix last line indentation
 
-  const names = [eventName, ...previousEventNames]
+  const names = [eventName, ...(previousEventNames || [])]
     .map((s) => JSON.stringify(s))
     .join(" | ");
 
--- a/assets/js/analytics-events.js
+++ b/assets/js/analytics-events.js
@@ -70,7 +70,11 @@ function Anchor({ slug, icon = String.fromCharCode(59851) }) {
   ></a>`;
 }
 
-function Example({ event_name: eventName, attributes }) {
+function Example({
+  event_name: eventName,
+  previous_event_names: previousEventNames,
+  attributes,
+}) {
   function typeExample(type) {
     switch (type) {
       case "Boolean":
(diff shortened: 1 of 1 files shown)


assets/js/analytics-events.js
```diff
@@ -107,8 +111,12 @@ function Example({ event_name: eventName, attributes }) {
     .join("\n      ")
     .replace(/ {2}$/, ""); // fix last line indentation
 
+  const names = [eventName, ...previousEventNames]
```
aduth, line 114: Below, we use optional chaining for `previousEventNames`. Can it be `undefined` or `null`? If so, I would expect that to error here. If not, can we remove the optional chaining?

aduth (approved):
Looks like event name can escape bounds of code block. Can we make it scrollable?

image

assets/js/analytics-events.js
zachmargolis: Since it's a newer property, depending on what backend it's deployed against it could be undefined.

I though that `[...undefined]` would act like `[*nil]` in Ruby so it would be safe....but just confirmed that it's not, I'll add a fix

zachmargolis: 37a98adc2347dce0b5639d353c16bf1682a41463

zachmargolis:
> Looks like event name can escape bounds of code block. Can we make it scrollable?

In that example, it's a specific property `type` that's breaking out not the name, but yes good point, fixed in 098bd18b0577472d26e8e9950134e8adeccdcdaf

assets/js/analytics-events.js
aduth: > I though that `[...undefined]` would act like `[*nil]` in Ruby so it would be safe....but just confirmed that it's not, I'll add a fix

🙈  https://www.typescriptlang.org/play?#code/MYewdgzgLgBADgJwKYDcCWICuECiKlhQByAhgLZIQwC8MYmANgwNwBQoksY5lNMA2gDphiVBmx4CxHhAC6zIA

stevegsa (approved):

PR #228, opened 2022-03-16
Update glossary: GPO and SP

zachmargolis:


mitchellhenke (approved):

_articles/glossary.md
```diff
@@ -226,8 +231,8 @@ A trusted entity that establishes and vouches for the identity or attributes of
 ### SAML
 **Security Assertion Markup Language**
 
-### Service Provider
-**SP**
+### SP
+**Service Provider**
 An application (or applications) within an ecosystem that provides a service to users; examples of SPs that relate to Login.gov include vets.gov, myssa, and ELIS.
```
aduth, line 236: We seem to be inconsistent with acronyms whether there's a line break between the expanded form and its description. I would think we want it with the line break?

```suggestion
**Service Provider**

An application (or applications) within an ecosystem that provides a service to users; examples of SPs that relate to Login.gov include vets.gov, myssa, and ELIS.
```

aduth (approved):
👍

_articles/glossary.md
zachmargolis: whoops, great catch

PR #233, opened 2022-03-22
modulify other JS on the page

zachmargolis:


aduth:
Might want to try to find a way to share common code across bundles to avoid duplicating vendor scripts, either with [ESBuild splitting](https://esbuild.github.io/api/#splitting) (see caveats) or not passing `--bundle` flag and allowing for ESM URL loading.

zachmargolis:
> Might want to try to find a way to share common code across bundles to avoid duplicating vendor scripts, either with [ESBuild splitting](https://esbuild.github.io/api/#splitting) (see caveats) or not passing `--bundle` flag and allowing for ESM URL loading.

I don't feel that strongly either way, do you have a preference for one approach or the other?

aduth:
I think `--splitting` will probably be easier to deal with external dependencies, assuming the caveats don't hurt us.

I just tested it real quick on this branch and it seems to work exactly how I'd expect ⭐ 

```
$ esbuild `find assets/js -type f -not -name '*.min.*'` --outdir=assets/build --bundle --format=esm --splitting "--minify"

  assets/build/analytics-events.js  11.4kb
  assets/build/chunk-T6EOKNVV.js     5.5kb
  assets/build/private-links.js      2.0kb
  assets/build/chunk-NLMHZ7JJ.js     544b 
  assets/build/anchors.js            214b 
```

```js
// assets/build/analytics-events.js
import {
  require_anchor
} from "./chunk-YCTBOAM6.js";
// ...
```

```js
// assets/build/anchors.js
import {
  require_anchor
} from "./chunk-YCTBOAM6.js";
// ...
```

aduth:
> I think `--splitting` will probably be easier to deal with external dependencies

[Import maps](https://github.com/WICG/import-maps) would be a nice solution to this, but [browser support](https://caniuse.com/import-maps) is pretty limited (Chromium-only) without a [polyfill](https://github.com/guybedford/es-module-shims).

aduth:
Re: Failing build, I think you ran into what I encountered with https://github.com/18F/identity-handbook/pull/232#discussion_r832542037 .

Quick-but-unideal solution is to remove the DefinitelyTyped package.

Alternatively, we could try to find a different way to enable the anchor features (via `Anchor#add`?), or override the type definitions like what we've done in the IdP ([example](https://github.com/18F/identity-idp/blob/main/typings/clipboard-polyfill/overwrite-globals.d.ts)).

zachmargolis:
> Alternatively, we could try to find a different way to enable the anchor features (via `Anchor#add`?), or override the type definitions like what we've done in the IdP ([example](https://github.com/18F/identity-idp/blob/main/typings/clipboard-polyfill/overwrite-globals.d.ts), [18F/identity-idp#5977](https://github.com/18F/identity-idp/pull/5977)).

I tried overriding the typings, with not much success in 2cae4c1 so I took the coward's way out and added `as any` 😬

aduth:
> > Alternatively, we could try to find a different way to enable the anchor features (via `Anchor#add`?), or override the type definitions like what we've done in the IdP ([example](https://github.com/18F/identity-idp/blob/main/typings/clipboard-polyfill/overwrite-globals.d.ts), [18F/identity-idp#5977](https://github.com/18F/identity-idp/pull/5977)).
> 
> I tried overriding the typings, with not much success in [2cae4c1](https://github.com/18F/identity-handbook/commit/2cae4c1d62adf895787ccf3811b6d844b6ce857c) so I took the coward's way out and added `as any` 😬

I was curious, and think I got it after some trial-and-error in [b7d79d0](https://github.com/18F/identity-handbook/pull/233/commits/b7d79d0251aac3880f6d77c0fcc073cf3516b7fb). If we needed the full types, we could also essentially copy what's in `@types/anchor-js` and add the `urlify` function declaration as part of [its `interface Anchor`](https://unpkg.com/browse/@types/anchor-js@4.2.1/index.d.ts).

aduth (approved):

aduth:
> If we needed the full types

Ah, based on build failure, we'll at least need a little bit more of the types:

```
assets/js/anchors.tsx:4:16 - error TS2339: Property 'add' does not exist on type 'Anchor'.
```

PR #235, opened 2022-03-22
Add more types

zachmargolis:
current issues:
- Can't seem to find/locate the types for 18f/private-eye that I added (is the slash a namespace issue?)
- simple-jekyll-search exports [by defining `window.SimpleJekyllSearch=`](https://github.com/christian-fei/Simple-Jekyll-Search/blob/8a9767f2d12652528c95b94176c4331c9df69b01/src/index.js#L42)... so I tried a stateful import + a separate definition of global variables


_layouts/default.html
```diff
@@ -74,23 +74,20 @@ <h3 class="usa-footer__logo-heading">U.S. General Services Administration</h3>
...
-        searchInput: document.getElementById('search-input'),
-        resultsContainer: document.getElementById('search-results'),
-        json: '{{ site.baseurl }}/search.json',
-        noResultsText: '<li class="no-results">No results were found.</li>'
-      });
-    </script>
     <script type="module">
       import { loadAnchors } from '{{ "/assets/build/anchors.js" | prepend: site.baseurl }}';
       import { loadPrivateEye } from '{{ "/assets/build/private-links.js" | prepend: site.baseurl }}';
+      import { loadSimpleJekyllSearch } from '{{ "/assets/build/search.js" | prepend: site.baseurl }}';
```
zachmargolis, line 80: another idea might be to put all these `loadXYZ` functions in a single `.tsx` and then we save a few requests, because the compiler might chunk them better?

assets/js/analytics-events.tsx
```diff
@@ -155,19 +178,25 @@ function Event({
   );
 }
 
-function Events({ events }) {
-  return events.map((event) => <Event {...event} />);
+function Events({ events }: { events: AnalyticsEvent[] }) {
+  return (
+    <>
```
zachmargolis, line 183: types got mad that `Element[]` was not an `Element` 🤷

_layouts/default.html
zachmargolis: done: 706690fa8a94f6033a1f04946b4907d908493043

_layouts/default.html
```diff
@@ -74,23 +74,22 @@ <h3 class="usa-footer__logo-heading">U.S. General Services Administration</h3>
...
 
       document.addEventListener('DOMContentLoaded', () => {
         loadAnchors();
         loadPrivateEye();
+        loadSimpleJekyllSearch({
+          searchInput: document.getElementById('search-input'),
+          resultsContainer: document.getElementById('search-results'),
+          json: '{{ site.baseurl }}/search.json',
+          noResultsText: '<li class="no-results">No results were found.</li>'
+        })
```
aduth, line 92: Nits:

```suggestion
          noResultsText: '<li class="no-results">No results were found.</li>',
        });
```

assets/js/setup.tsx
```diff
@@ -0,0 +1,65 @@
+import Anchor from "anchor-js";
```
aduth, line 1: I can't tell if this file is supposed to be DOM-aware or unaware. For some things we expect options to be passed in (`loadSimpleJekyllSearch` elements), and we don't handle the initialization here, but in other cases we specify selectors directly (`loadAnchors` selectors).

Wonder if we can be consistent one way or the other:

- Do everything in this script, including initialization
- Be consistent with initialization options being passed in

I probably lean a little more toward the first, since:

- Linting doesn't run on inline code
- We can get rid of the `load*` functions and just call the constructors inline

assets/js/typings/18f-private-eye/index.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module "@18f/private-eye" {
```
aduth, line 1: I think to get the namespacing to work, the folder would be named `18f__private-eye`.

Ref: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master#what-about-scoped-packages

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "strict": true,
-    "noImplicitAny": false,
+    "noImplicitAny": true,
```
aduth, line 4: `noImplicitAny` defaults `true` if `strict` is `true`. It was only necessary to explicitly override it to `false`, but we can remove it altogether if it's `true` now.

```suggestion
```

Ref: https://www.typescriptlang.org/tsconfig#noImplicitAny

aduth (approved):
👍

assets/js/setup.tsx
zachmargolis: I wanted to make sure that SimpleJekyllSearch's `json` parameter was built in the template so that we didn't have to worry about `{{ site.baseurl }}` inside of a JS file.

otherwise, I agree, would be happy to do everything in this file. Maybe we update it so that that's the only option passed in, and the rest are hard coded in here?

or we could put it as a `data-` attribute, maybe on the search box?

aduth: > I wanted to make sure that SimpleJekyllSearch's `json` parameter was built in the template so that we didn't have to worry about `{{ site.baseurl }}` inside of a JS file.

Ah yeah, forgot about the `{{ site.baseurl }}` 🤔 

But to your other point, yeah, we could put it on an attribute somewhere, like `html`, `body`, the search box, or just an empty script/noscript tag ([see similar in IdP](https://github.com/18F/identity-idp/blob/0b44685a9d6ad77ad3eba84c6f55b65542476fc8/app/views/layouts/base.html.erb#L87)).

assets/js/typings/18f-private-eye/index.d.ts
zachmargolis: I tried it in c636bd31c86d757289dbbd4e4708ae6381e8add4 and it doesn't seem to have fixed the issue

zachmargolis: But I did see https://github.com/18F/private-eye/pull/50! Hoping that will be easier to work with

assets/js/setup.tsx
```diff
@@ -0,0 +1,60 @@
...
+import PrivateEye from "@18f/private-eye";
+import "simple-jekyll-search";
+
+// eslint-disable-next-line import/no-unresolved
+import { SimpleJekyllSearchGlobals } from "simple-jekyll-search-types";
```
aduth, line 6: Does it make a difference if we use `import type` here?

```suggestion
import type { SimpleJekyllSearchGlobals } from "simple-jekyll-search-types";
```

zachmargolis: oh wow, TIL about `import type` --- thanks! PR to fix coming up shortly

zachmargolis: https://github.com/18F/identity-handbook/pull/237

PR #236, opened 2022-03-22
Co-opt Netlify login for private content on handbook

zachmargolis:
- Is the user experience smooth? No
- Is this a well-designed experience? Also no
- Does it work? Barely

Next up: leveraging the private handbook repo and injecting private articles 

| | logged out | logged in |
| ---- | ---- | ---- |
| login status | Screen Shot 2022-03-22 at 4 06 56 PM | Screen Shot 2022-03-22 at 4 06 20 PM |
| specific article content | Screen Shot 2022-03-22 at 4 07 14 PM | Screen Shot 2022-03-22 at 4 07 51 PM |


_layouts/default.html
```diff
@@ -88,46 +90,21 @@ <h3 class="usa-footer__logo-heading">U.S. General Services Administration</h3>
...
+        loadAnchors();
+        loadPrivateEye();
+      }, false);
+    </script>
+    <script type="module">
+      import { setUpPrivate } from '{{ "/assets/build/private.js" | prepend: site.baseurl }}';
+
+      setUpPrivate({
+        buttonContainer: document.getElementById('private-login-button-container'),
+      })
```
aduth, line 107: Nit:

```suggestion
      });
```

assets/js/private.tsx
```diff
@@ -0,0 +1,69 @@
...
+          height="30"
+        />
+        @{currentUser.login}
+        <button
+          type="button"
+          className="usa-button usa-button--gray"
+          onClick={() => {
+            logOut();
+            setCurrentUser(undefined);
+          }}
```
aduth, line 49: It might be nice if the state were more tightly coupled to the localStorage value so that we'd just need to call one function here.

```suggestion
          onClick={logOut}
```

Couple ideas:

- Could move `logOut` inside the component and have it call `setCurrentUser`
- Could create a custom hook to manage localStorage + state, e.g. `const [user, logOut] = useUser();`

assets/js/private.tsx
```diff
@@ -0,0 +1,69 @@
...
+        >
+          Log Out
+        </button>
+      </div>
+    );
+  }
+  return (
+    <a className="usa-button" href="/admin">
```
aduth, line 57: Should we try to make this work in Federalist?

aduth: The styling looks kinda busted in the preview:

image

Looks like `usa-nav__secondary-links` color is taking precedent.

Couple thoughts:

- Could we move the button elsewhere, perhaps either adjacent the "Edit" button at the bottom of the page, or contextual to whichever content would require it (e.g. "Log in to view this content").
- Inline style tag to set the color to white (Ideally not 😅 )

zachmargolis: I'll add `<body data-base-url={{site.baseurl}}>` so that way we can use that in a bunch of places to build the right admin path

zachmargolis: I don't have a very strong opinion on where the button goes, I thought the top made it easier to find

_layouts/default.html
```diff
@@ -88,46 +90,21 @@ <h3 class="usa-footer__logo-heading">U.S. General Services Administration</h3>
...
+    <script type="module">
+      import { loadAnchors } from '{{ "/assets/build/anchors.js" | prepend: site.baseurl }}';
+      import { loadPrivateEye } from '{{ "/assets/build/private-links.js" | prepend: site.baseurl }}';
+
+      document.addEventListener('DOMContentLoaded', () => {
+        loadAnchors();
+        loadPrivateEye();
+      }, false);
+    </script>
+    <script type="module">
```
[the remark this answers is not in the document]
zachmargolis: updated in 3eb99034de1e36060ac7facce165693466ae2fc5

assets/js/analytics-events.tsx
```diff
@@ -192,23 +193,38 @@ function SidebarNavItem({ name }: { name: string }) {
   );
 }
 
-function ErrorPage({ error, url }: { error: Error; url: string }) {
+function AlertComponent({
```
aduth, line 196: Maybe drop the "Component" suffix, since it's assumed.

```suggestion
function Alert({
```

assets/js/analytics-events.tsx
```diff
@@ -192,23 +193,38 @@ function SidebarNavItem({ name }: { name: string }) {
...
-        </div>
+        <h5 className="usa-alert__heading">{heading}</h5>
+        <div className="usa-alert__text">{content}</div>
       </div>
     </div>
   );
 }
 
+function ErrorPage({ error, url }: { error: Error; url: string }) {
+  return AlertComponent({
```
aduth, line 214: Should we be calling this as a component?

```suggestion
  return <AlertComponent
```

assets/js/analytics-events.tsx
```diff
@@ -192,23 +193,38 @@ function SidebarNavItem({ name }: { name: string }) {
...
+    heading: "Error loading event definitions",
+    content: (
```
aduth, line 216: As a component, I'd see this as children.

Related: https://github.com/18F/identity-idp/blob/main/app/javascript/packages/components/alert.jsx

assets/js/analytics-events.tsx
```diff
@@ -221,28 +237,39 @@ function Sidenav({ events }: { events: AnalyticsEvent[] }) {
...
+        }
+      })
+      .catch((error) =>
+        render(<ErrorPage url={eventsUrl} error={error} />, container)
+      );
+  } else {
+    render(
+      <AlertComponent
+        heading="Error loading event definitions"
+        content="Please log in"
```
aduth, line 270: Maybe have the button here?

Or at least in addition to header?

assets/js/analytics-events.tsx
zachmargolis: 🤦 wow yes that makes more sense

assets/js/analytics-events.tsx
zachmargolis: Gave it a shot like this? It is not pretty

Screen Shot 2022-03-23 at 12 53 15 PM

assets/js/private.tsx
```diff
@@ -0,0 +1,79 @@
...
+      }
+    } catch {}
+  }
+}
+
+function removeLoggedInUser() {
+  localStorage.removeItem(NETLIFY_USER_KEY);
+}
+
+function useCurrentUser(): [NetlifyUser | undefined, () => void] {
```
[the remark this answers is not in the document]
zachmargolis: I like that idea --- I am not sure how to keep state shared across two separately rendered start points? Will leave that task to future us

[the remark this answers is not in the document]
aduth: Hm, yeah, a bit tricky. I was wondering if there's some kind of event from `localStorage` manipulation. There's the [`storage` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event), but it doesn't look like it works within the same page.

It's probably something where we can use context, keep track of subscribers, and notify them when an update has happened, similar to what we do in the [IdP's `useCookie`](https://github.com/18F/identity-idp/blob/main/app/javascript/packages/document-capture/hooks/use-cookie.js).

zachmargolis:
Updated with new link styles:

| screen | screenshot |
| --- | --- |
| top corner | Screen Shot 2022-03-23 at 1 57 51 PM |
| inline on events page |  Screen Shot 2022-03-23 at 1 57 59 PM |

aduth:
Admittedly I don't understand why the secondary links are styled as the same color as text, but I see it's coming from the design system, so... 🤷  Upstream concern.

aduth (approved):

PR #238, opened 2022-03-28
Add dynamically loaded team page from GitHub

zachmargolis:
- We can drop this page from the old private GitHub now
- Working example of pulling private content for our team from GitHub. Next up: entirely private articles

| logged out | logged in |
| --- | --- |
| logged-out | logged-in |

### DIFF
--- a/_articles/team.md
+++ b/_articles/team.md
@@ -1,0 +1,28 @@
+---
+title: Team Members
+description: The login.gov team roster
+layout: article
+category: Team
+---
+
+This page is dynamically generated from [identity-private/team.yml][team-yml]
+every time the handbook is deployed.
+
+[team-yml]: https://github.com/18F/identity-private/blob/main/team/team.yml
+
+<div id="error-container"></div>
+
+### Current Team Members
+
+<div id="team-container"></div>
+
+### Alumni
+
+<div id="alumni-container"></div>
+
+<script type="module">
+import { loadTeam } from '{{ "/assets/build/team.js" | prepend: site.baseurl }}';
+
+loadTeam();
+</script>
+
--- a/assets/js/components/alert.tsx
+++ b/assets/js/components/alert.tsx
@@ -1,0 +1,18 @@
+import { h, ComponentChildren } from "preact";
+
+export function Alert({
+  heading,
+  children,
+}: {
+  heading?: string;
+  children: ComponentChildren;
+}) {
+  return (
+    <div className="usa-alert usa-alert--error">
+      <div className="usa-alert__body">
+        <h5 className="usa-alert__heading">{heading}</h5>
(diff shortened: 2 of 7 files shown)


_articles/team.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Team Members
+description: The login.gov team roster
+layout: article
+category: Team
+---
+
+This page is dynamically generated from [identity-private/team.yml][team-yml]
+every time the handbook is deployed.
```
zachmargolis, line 9:
```suggestion
```

assets/js/components/alert.tsx
```diff
@@ -0,0 +1,18 @@
+import { h, ComponentChildren } from "preact";
+
+export function Alert({
+  heading,
+  children,
+}: {
+  heading?: string;
```
aduth, line 7: It existed previously, and we're not currently leveraging it's optional-ness, but I'm thinking we should either make this required, or handle it in the render logic so that we don't output the `<h5>` if it's empty, in order to avoid (1) weird visual offset and (2) [accessibility issue](https://dequeuniversity.com/rules/axe/3.5/empty-heading?application=AxeChrome). That said, ideally we aren't implementing these components in each project, either having one source of design system components, or using something like [react-uswds](https://github.com/trussworks/react-uswds).

image

```suggestion
  heading: string;
```

assets/js/team.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return <a href={`https://github.com/${username}`}>@{username}</a>;
+}
+
+function TeamMemberRoster({ members }: { members: TeamMember[] }) {
+  return (
+    <ul>
+      {members
+        .sort(({ name: aName }, { name: bName }) => aName.localeCompare(bName))
+        .map(({ name, github, email, subteam, responsibilities, fed, aws }) => (
+          <li>
```
aduth, line 42: Preact isn't as loud about it as React, but `key`s are still a good practice.

https://preactjs.com/tutorial/08-keys/#the-key-to-list-rendering
https://reactjs.org/docs/lists-and-keys.html#keys

Could be relevant if ever we'd allow sorting or re-updating of the list.

```suggestion
          <li key={email}>
```

aduth (approved):
Nice! 👍

assets/js/components/alert.tsx
zachmargolis: updated component to avoid the empty tag in d56d337153c68f75b078bd4d374afe14a2eca2c5

assets/js/team.tsx
zachmargolis: fab1062797b609c8c637781b47b942927ed6e75a

PR #241, opened 2022-03-29
Redo team page components

zachmargolis:
- Leverage useQuery and createPortal

PR #242, opened 2022-04-04
Update preact-fetching, also fix response destructuring

zachmargolis:
Two separate things, I thought I already fixed the first one but apparently not

1. Update `preact-fetching` so we can destructure `data` and `error` in the same line
2. Fix the structure of destructuring analytics events JSON, I got it wrong in #241


aduth (approved):
LGTM 👍

PR #243, opened 2022-04-08
Better descibe Scribe role

pauldoomgov:
Refines Scribe role description.   Explicitly removes expectation
of a full transcript.


aduth (approved):
👍

PR #248, opened 2022-04-28
add final step to appdev deploying handbook page

mitchellhenke:
There's some ambiguity in going from one section to the next, so this adds a final step to add a bit of separation and indicate when a deployer is done deploying 🙂


aduth (approved):
Makes sense! 👍

PR #250, opened 2022-05-09
Update guidance around JS internationalization

aduth:
**Why**: Because the current guidance is outdated, referencing an allowlist file and global function which no longer exist.

h/t @jc-gsa for surfacing

### DIFF
--- a/_articles/appdev-i18n.md
+++ b/_articles/appdev-i18n.md
@@ -37,14 +37,12 @@ t('doc_auth.buttons.start_over')
 
 ### Front-end
 
-Login.gov uses a custom JavaScript implementation providing basic support for a Rails-like `t` translation function.
+Login.gov uses a [custom JavaScript implementation](https://github.com/18F/identity-idp/tree/main/app/javascript/packages/i18n) which provides basic support for a Rails-like `t` translation function.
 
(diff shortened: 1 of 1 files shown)


_articles/appdev-i18n.md
```diff
@@ -37,14 +37,12 @@ t('doc_auth.buttons.start_over')
 
 ### Front-end
 
-Login.gov uses a custom JavaScript implementation providing basic support for a Rails-like `t` translation function.
+Login.gov uses a [custom JavaScript implementation](https://github.com/18F/identity-idp/tree/main/app/javascript/packages/i18n) which provides basic support for a Rails-like `t` translation function.
```
aduth, line 40: The link destination isn't very helpful in its current form, but it would be nice to add a README.md file there, which would show automatically at the link destination if/when implemented.

ctro (approved):
LGTM

jc-gsa (approved):

_articles/appdev-i18n.md
aduth: > The link destination isn't very helpful in its current form, but it would be nice to add a README.md file there, which would show automatically at the link destination if/when implemented.

Update: The link is now useful!

PR #256, opened 2022-05-12
Update glossary.md

benjaminchait:
I think these two lines (96 and 97) are missing the closing `_` for that text to be _italicized_? 

But I could be mistaken, feel free to let me know by rejecting/commenting!

As it stands, this section is rendered as such: 
image


zachmargolis (approved):
good catch! LGTM

PR #259, opened 2022-05-23
Bump PrivateEye dependency for first-party type definitions

aduth:
**Why**: So that we don't need to maintain them ourselves.

Related:

- https://github.com/18F/private-eye/pull/50
- #235

### DIFF
--- a/assets/js/private-links.tsx
+++ b/assets/js/private-links.tsx
@@ -1,7 +1,8 @@
 import PrivateEye from "@18f/private-eye";
 
 export const loadPrivateEye = () => {
-  PrivateEye({
+  // eslint-disable-next-line no-new
+  new PrivateEye({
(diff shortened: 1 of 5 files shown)


assets/js/private-links.tsx
```diff
@@ -1,7 +1,8 @@
 import PrivateEye from "@18f/private-eye";
 
 export const loadPrivateEye = () => {
-  PrivateEye({
+  // eslint-disable-next-line no-new
```
aduth, line 4: Not ideal, but the constructor call form is the canonical recommended usage, per the code source.

https://github.com/18F/private-eye/blob/v2.1.0/private-eye.js#L48-L50

zachmargolis (approved):
LGTM

PR #262, opened 2022-05-24
Update definition-of-done.md

benjaminchait:
Add link to #login-ops Slack channel.

---

I noticed these were plain text but didn't see the (private) channel. @ctro confirmed this is real and added me, I _think_ making these links might help in the future. 

@zachmargolis - Is this formatted properly, to get the 🔒 beside a Slack private link?

### DIFF
--- a/_articles/definition-of-done.md
+++ b/_articles/definition-of-done.md
@@ -24,9 +24,9 @@ Expect this DoD to change over time.
 - If multiple database migrations are necessary, the database migrations have been performed
 - If removing a previously populated billing or accounting related item from the database (changing `sp_costs`):
   - Ensure there is a Jira ticket to link the change to and reference it in the PR
-  - Drop a line in #login-ops noting the proposed change and ensure the data is not needed for billing/reporting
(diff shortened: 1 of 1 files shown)


benjaminchait:
@zachmargolis - Also, I'm not sure why, but the `build` doesn't seem to be working from my user? (Last time, @aduth had to "bump circle" to get it to build properly? see: https://github.com/18F/identity-handbook/pull/256)

benjaminchait:
@zachmargolis @aduth - To recap, my user had never signed into circleci.com and thus none of the builds were triggering from my PRs. Making a new commit after auth'ing to circleci seems to have the build working, and I'll merge when complete. Thanks again!

aduth:
> @zachmargolis @aduth - To recap, my user had never signed into circleci.com and thus none of the builds were triggering from my PRs. Making a new commit after auth'ing to circleci seems to have the build working, and I'll merge when complete. Thanks again!

That's good to know! I didn't realize someone has to sign in to CircleCI for builds to automatically trigger 💡

PR #264, opened 2022-05-25
Update glossary.md

benjaminchait:
@zachmargolis - Taking our Slack DM and putting some words behind it, but feel free to edit/wordsmith as you see fit.


zachmargolis (approved):
LGTM

PR #269, opened 2022-06-09
Remove sudo from appdev deploy step commands

aduth:
**Why**: Because you no longer need to (and cannot) use sudo, so the example command may cause some confusion in its current form.


zachmargolis (approved):
LGTM

PR #274, opened 2022-06-28
Update On/offboarding to remove mention of team.yml

ctro:
Remove `team.yml` guidance


aduth (approved):
Is the plan to remove this file? And if so, should we also plan to remove [the "Team Members" handbook article](https://handbook.login.gov/articles/team.html)?

ctro:
> Is the plan to remove this file? And if so, should we also plan to remove [the "Team Members" handbook article](https://handbook.login.gov/articles/team.html)?

@aduth `team.yml` is out of date, and we can consider it deprecated.  I'm not sure if we should remove from it's own repo -- there seems to be a lot of other cruft in that repo.  Also yes we should remove the "team" page from the handbook that reads `team.yml`.  Do I need to remove anything other than the article?  Seems like maybe there is some JS someplace.

aduth:
Yeah I think removing these two files should cover it:

- `_articles/team.md`
- `assets/js/team.tsx`

ctro:
Yeah, sad to delete all that beautiful code :(

PR #279, opened 2022-07-07
Redo the "troubleshooting production" article

zachmargolis:
- Include SSM documents
- Include Cloudwatch tips


zachmargolis:
> Looks good! For the saved query screenshot it might be nice to highlight one of our Terraformed per-environment searches like:

Wow I didn't even know we had all those! Updated in 8e142152ac8aec76da9015347585e7be467683e2

PR #280, opened 2022-07-07
Normalize sprint team data to a .yml file

zachmargolis:
This might be a little less accessible to maintain, so I am open to just discarding it, but it seems like the more teams we onboard, the easier it would be to track various fields and make sure they get formatted properly?

### DIFF
--- a/_articles/slack.md
+++ b/_articles/slack.md
@@ -41,7 +41,7 @@ These handles ping oncall engineers, use for emergencies or urgent items.
 - `@login-feds` Group for Login.gov federal FTE employees
 - `@github-admins-login` members of this group can edit [`identity-core`](https://github.com/orgs/18F/teams/identity-core) GitHub team membership
 
-#### Sprint Teams
+### Sprint Team Handles
 
 {% comment %}
 see data/sprint_teams.yml for the data
@@ -49,11 +49,9 @@ see data/sprint_teams.yml for the data
 
 {% assign sprint_team_handles = site.data.sprint_teams | sort: "name" %}
 
-{% for sprint_team in sprint_team_handles %}
-{% for slack_group_name in sprint_team.slack_group_names %}
-- `@{{ slack_group_name }}` To tag members of [{{ sprint_team.name }}]({% link _articles/sprint-teams.md %}#{{ sprint_team.name | slugify }})
-{%- endfor -%}
-{%- endfor %}
+{% for sprint_team in sprint_team_handles %}{% if sprint_team.slack_group_name %}
+- `@{{ sprint_team.slack_group_name }}` To tag members of [{{ sprint_team.name }}]({% link _articles/sprint-teams.md %}#{{ sprint_team.name | slugify }})
+{%- endif %}{%- endfor %}
 
 
 ## Channels
--- a/_articles/slack.md
+++ b/_articles/slack.md
@@ -117,20 +117,6 @@ Here are some common Slack Channels for the Login.gov team
 - [`#login-reporting`](https://gsa-tts.slack.com/archives/C5E7EJWF7):
   Analytics and reporting channel
 
-{% comment %}
-see data/sprint_teams.yml for the data
-{% endcomment %}
-
-{% assign sprint_team_channels = site.data.sprint_teams | sort: "name" %}
-
-{% for sprint_team in sprint_team_channels %}
-{% if sprint_team.slack_channel_name and sprint_team.slack_channel_url %}
-- [`#{{ sprint_team.slack_channel_name }}`]({{ sprint_team.slack_channel_url }})
-  {% include tag.html name='appdev' color='bg-blue' %}
-  Team {{ sprint_team.name }} channel
-{% endif %}
-{% endfor %}
-
 #### Engineering/DevOps
 
 - [`#login-devops`](https://gsa-tts.slack.com/archives/C16RSBG49):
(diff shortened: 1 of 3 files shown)


_data/sprint_teams.yml
```diff
@@ -0,0 +1,71 @@
...
+  - login-joy-designers
+
+- name: Judy
+  namesake_markdown: "[Judy Parsons](https://blog.theveteranssite.greatergood.com/wwii-navy-codebreaker/), As a code breaker working with a secret unit in the US Navy, Parsons was instrumental in deciphering the enemy’s secret codes, but her work went unacknowledged for decades after the war."
+  focus_area: Cybersecurity and Anti-Fraud work stream
+  slack_channel_url: https://gsa-tts.slack.com/archives/C03A12WPCLW
+  slack_channel_name: login-team-judy
+
+- name: Kimberly
+  namesake_markdown: "[Kimberly Bryant](https://en.wikipedia.org/wiki/Kimberly_Bryant_(technologist), an electrical engineer who founded [Black Girls Code](https://www.blackgirlscode.com/)."
```
aduth, line 44: The parenthesis in the link seems to be messing with the rendering. May just need another one to close it out?

image

```suggestion
  namesake_markdown: "[Kimberly Bryant](https://en.wikipedia.org/wiki/Kimberly_Bryant_(technologist)), an electrical engineer who founded [Black Girls Code](https://www.blackgirlscode.com/)."
```

_data/sprint_teams.yml
```diff
@@ -0,0 +1,71 @@
...
+  slack_channel_url: https://gsa-tts.slack.com/archives/C017KUF5DL4
+
+- name: Joy
+  namesake_markdown: "[Dr. Joy Buolamwini](https://en.wikipedia.org/wiki/Joy_Buolamwini), a computer scientist, artist, and activist who has published foundational work on algorithmic discrimination."
+  focus_area: In Person Proofing
+  slack_channel_name: login-team-joy
+  slack_channel_url: https://gsa-tts.slack.com/archives/C03FA4VBN76
+  slack_group_names:
+  - login-joy-engineers
+  - login-joy-designers
```
aduth, line 35: Should we include the broader team handle as well?

```suggestion
  slack_group_names:
  - login-joy
  - login-joy-engineers
  - login-joy-designers
```

_articles/slack.md
```diff
@@ -104,24 +116,22 @@ Here are some common Slack Channels for the Login.gov team
...
-  {% include tag.html name='appdev' color='bg-blue' %}
-  Team Grace channel
+{% comment %}
+see data/sprint_teams.yml for the data
+{% endcomment %}
 
-- [`#login-katherine`](https://gsa-tts.slack.com/archives/C01710KMYUB):
-  {% include tag.html name='appdev' color='bg-blue' %}
-  Team Katherine channel
+{% assign sprint_team_channels = site.data.sprint_teams | sort: "name" %}
```
aduth, line 126: I do like how having a single source of truth for the data lets us list the Slack channels here.

But should they all be under "AppDev" ? I would think of teams like "Radia" as under "Engineer/DevOps", and maybe "Ursula" under "Partners".

Should we include Slack category in the data? Or, should we create a separate section in the list of Slack channels for team channels, and move all of them there?

_articles/slack.md
```diff
@@ -39,11 +39,23 @@ These handles ping oncall engineers, use for emergencies or urgent items.
...
+
+{% comment %}
+see data/sprint_teams.yml for the data
+{% endcomment %}
+
+{% assign sprint_team_handles = site.data.sprint_teams | sort: "name" %}
+
+{% for sprint_team in sprint_team_handles %}
+{% for slack_group_name in sprint_team.slack_group_names %}
+- `@{{ slack_group_name }}` To tag members of [{{ sprint_team.name }}]({% link _articles/sprint-teams.md %}#{{ sprint_team.name | slugify }})
```
aduth, line 54: It's a bit odd that Joy's engineer/designer-specific groups are listed twice as a way to "tag members of Joy".

image

Per my prior comment, maybe we should _only_ list the broader sprint team group for `login-joy` ?

aduth, line 54: Not sure if there's a way to fix it, but there are some paragraph tags sneaking in for some (but not all) of the team listings, creating some inconsistent spacing:

image

aduth (approved):
Seems like a good idea! I'd like to think the standard structure of the YAML file should be straight-forward enough for folks to copy-paste and edit. I also like how it allows for us to be consistent across multiple articles referencing the sprint teams.

_articles/slack.md
zachmargolis: I had a brainfart and forgot that plain `login-joy` existed... I will switch this to be only `login-joy` which means this can be a single item (not an array), which means I can fix it all:
- include the whole team tag
- fix the spacing because there's one fewer loop to worry about leading/trailing whitespac in
- remove redundant text

fixed it all in b05fb04d707555f4f952ee548e91a1854bf022e1

_data/sprint_teams.yml
zachmargolis: went with singular slack handle approach in b05fb04d707555f4f952ee548e91a1854bf022e1

_articles/slack.md
zachmargolis: moved to separate section in d64c8f935a18c6f96190ca9fd5ff17082f662638

PR #285, opened 2022-07-20
simplified gitlab env deploy config

timothy-spencer:
https://github.com/18F/identity-devops/pull/4956 removed the need for us to duplicate config in chef and tf, so trueing up the docs to match that.


zachmargolis (approved):
LGTM yay simplification

PR #292, opened 2022-08-18
Add Slack links to sprint-teams.md

benjaminchait:
I see in `main/_data/sprint_teams.yml` that we have both a `slack_channel_name` and `slack_channel_url` as variables, but I don't see them on our Sprint Teams page. Maybe that's intentional, but it seems helpful if we point people to this Handbook page?


zachmargolis (approved):
LGTM! Makes sense

PR #295, opened 2022-08-19
Remove references to idpxtra from deploy guide

aduth:
**Why**: Since it no longer exists

### DIFF
--- a/_articles/appdev-deploy.md
+++ b/_articles/appdev-deploy.md
@@ -269,4 +269,5 @@ new configurations (config from S3, or service provides from `identity-idp-confi
 
     ```bash
     aws-vault exec prod-power -- ./bin/scale-remove-old-instances prod idp
+    aws-vault exec prod-power -- ./bin/scale-remove-new-instances prod worker
     ```
--- a/_articles/appdev-deploy.md
+++ b/_articles/appdev-deploy.md
@@ -269,5 +269,5 @@ new configurations (config from S3, or service provides from `identity-idp-confi
 
     ```bash
     aws-vault exec prod-power -- ./bin/scale-remove-old-instances prod idp
-    aws-vault exec prod-power -- ./bin/scale-remove-new-instances prod worker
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM ❤️

_articles/appdev-deploy.md
```diff
@@ -267,9 +265,8 @@ new configurations (config from S3, or service provides from `identity-idp-confi
    aws-vault exec prod-power -- ./bin/asg-recycle prod idp --skip-migration
    ```
 
-1. In production, it's important to remember to still scale out old IDP and IDPxtra instances.
+1. In production, it's important to remember to still scale out old IDP instances.
 
     ```bash
     aws-vault exec prod-power -- ./bin/scale-remove-old-instances prod idp
-    aws-vault exec prod-power -- ./bin/scale-remove-old-instances prod idpxtra
```
zachmargolis, line 274: should this section include workers?

aduth: If it's just a configuration change, we shouldn't need to recycle workers, do we?

zachmargolis: Yes, but since the instructions these days are to do a "full recycle" that would still end up recycling workers too

aduth: Turns out you're right! I'll add a worker command here.

_articles/appdev-deploy.md
```diff
@@ -267,9 +265,9 @@ new configurations (config from S3, or service provides from `identity-idp-confi
...
+    aws-vault exec prod-power -- ./bin/scale-remove-new-instances prod worker
```
zachmargolis, line 272: well it may not matter here but this is a different command than above
```suggestion
    aws-vault exec prod-power -- ./bin/scale-remove-old-instances prod worker
```

aduth: Whoops, thanks for catching that.

PR #301, opened 2022-08-31
Update help-center-contact-form.md

jmdembe:
The agency picker is not a freeform field, so that set of instructions needs to be removed.


aduth (approved):
LGTM 👍

zachmargolis:
😭 when did that change?

PR #304, opened 2022-09-01
create temporary netlify sandbox section

jmdembe:
I am looking to create a temporary sandbox section for NetlifyCMS demos. If there is a better place to put these, please let me know.


zachmargolis (approved):
LGTM

PR #306, opened 2022-09-06
Fix broken link

zachmargolis:
This is my guess at answering @aduth 's [question in slack](https://gsa-tts.slack.com/archives/CR6BF3CQK/p1662490664609689)


aduth (approved):
Looks reasonable to me! But I also lack the context of what had existed previously.

zachmargolis:
will merge now and we can always re-point later, yay source control

PR #308, opened 2022-09-06
Add a style for a highlighted target

zachmargolis:
- Used when linking to a particular anchor in the handbook

Example

Screen Shot 2022-09-06 at 1 06 23 PM

PR #309, opened 2022-09-07
Upgrade, use design system utilities for color reference

aduth:
Follow-up to TODO item in #308

- Upgrades `identity-style-guide` dependency from `6.3.0` to `6.5.0` to take advantage of latest bug fixes and improvements.
- Imports design system utility methods to avoid hard-coded reference to color value.
   - This project [uses the precompiled stylesheet](https://github.com/18F/identity-handbook/blob/dca78d9ed4649c741858bb63a09cb47bb44271e6/_config.yml#L16-L17) shipped through the design system, hence why utility methods are not available by default.

There should be no visual impact by these changes. Test to ensure that the behavior added in #308 continues to work.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -13,6 +13,10 @@ collections:
   articles: # individual articles
     output: true
 
+sass:
+  load_paths:
+    - node_modules
+
 copy_to_destination:
   - node_modules/identity-style-guide/dist/assets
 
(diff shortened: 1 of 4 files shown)


zachmargolis (approved):
lgtm, thanks for helping clean that up!

PR #310, opened 2022-09-08
Add anchors to links to external articles

zachmargolis:
**Why**: If we have a private google doc that's
findable from the handbook, I want to be able to link to the handbook for where to find that article

Example
- before, you could use Chrome's ["link to snippet"](https://handbook.login.gov/#:~:text=Environment%20Descriptions%0AListing%20of%20environments%20and%20the%20differences%20between%20them%2C%20like%20prod%2C%20pt%2C%20dm%2C%20int%20or%20dev)
- now, you get a little anchor UI element and a specific ID

| before | after |
| ---- | ---- |
|  Screen Shot 2022-09-08 at 3 25 58 PM | Screen Shot 2022-09-08 at 3 26 17 PM |

### DIFF
--- a/_includes/articles_category.md
+++ b/_includes/articles_category.md
@@ -17,10 +17,12 @@ includes:
 #### {{ subcategory }}
 
 {%     for article in articles %}
- - [{{article.title}}]({% include article_url.txt article=article %})
+ - <span markdown="1" {% if article.redirect_to %}data-anchor{% endif %}>
+   [{{article.title}}]({% include article_url.txt article=article %})
      {%- if article.deprecated -%}
        {: .deprecated-link} (deprecated)
      {%- endif -%}
+    </span>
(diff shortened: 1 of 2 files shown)


_includes/articles_category.md
```diff
@@ -17,10 +17,12 @@ includes:
 #### {{ subcategory }}
 
 {%     for article in articles %}
- - [{{article.title}}]({% include article_url.txt article=article %})
+ - <span markdown="1" {% if article.redirect_to %}data-anchor{% endif %}>
```
zachmargolis, line 20: `data-anchor` is basically a made-up attribute we use to say "hey, `anchor.js`, please add an anchor to this item":

https://github.com/18F/identity-handbook/blob/bef1b11803cc3eab92cb3d5a18823a997bf53161/assets/js/setup.tsx#L9

aduth (approved):
Nice 👍

PR #302, opened 2022-08-31
Update the deploy guide to describe 2x week deploys

jmhooper:
We talked about this at our staff eng weekly and at engineering huddle. This commit updates the documentation to describe that new process

PR #313, opened 2022-09-12
Update deployment documentation for twice-weekly deployments

aduth:
**Why:** So that the documentation is in sync with our current practices.

- Updates references to deployment days
- Clarifies that deployment is skipped for holidays, since it's more likely now with Monday deploys (e.g. Labor Day)
- Add optional step to use monitoring script for manual inspection

### DIFF
--- a/_articles/appdev-deploy.md
+++ b/_articles/appdev-deploy.md
@@ -14,7 +14,7 @@ A few notes on our deploy process.
 ### Cadence
 
 **When to deploy:** ✅
-- Typically we do a full deploy every week on Thursday.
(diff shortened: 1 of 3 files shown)


zachmargolis (approved):
LGTM, should we close out https://github.com/18F/identity-handbook/pull/302 since that's essentually a dupe of this?

aduth:
> LGTM, should we close out #302 since that's essentually a dupe of this?

🤦 

Despite the fact I reviewed it, I totally forgot about that other pull request. I think there's some nice unique bits from both so I'll coauthor a commit to merge some of @jmhooper's content into this, merge it, and close the other.

PR #315, opened 2022-09-14
Update documentation on app-s3-secret

zachmargolis:
(my action item from today's eng huddle)

I already updated [devops wiki page](https://github.com/18F/identity-devops/wiki/Common-Infrastructure-Commands-and-Shortcuts#app-s3-secret) to point to the exisitng page


pauldoomgov (approved):
Looks great!   Nice and simple.

PR #319, opened 2022-09-26
Update anchor link headings for dynamic articles

zachmargolis:
The dynamic headings in some articles don't play well with the `:target` styling added in #308 

**Note**: This doesn't seem to work on first load, so I'm putting it in draft. I'm not 100% sure why, maybe there's a load ordering or race condition issue somewhere

| before | after |
| --- | --- |
| Screen Shot 2022-09-26 at 1 35 40 PM | Screen Shot 2022-09-26 at 1 35 32 PM |

### DIFF
--- a/assets/js/analytics-events.tsx
+++ b/assets/js/analytics-events.tsx
@@ -102,18 +102,23 @@ function Event({ event }: { event: AnalyticsEvent }) {
     attributes = [],
   } = event;
 
+  const slug = urlify(eventName);
+  const previousEventSlug = urlify(`${eventName} Previous`);
+  const attributesSlug = urlify(`${eventName} Attributes`);
+  const exampleSlug = urlify(`${eventName} Example`);
+
   return (
     <div>
-      <h3>
+      <h3 id={slug}>
         {eventName}
-        <AnchorLink slug={urlify(eventName)} />
+        <AnchorLink slug={slug} />
       </h3>
       <p>{description}</p>
       {previousEventNames?.length ? (
         <>
-          <h4>
+          <h4 id={previousEventSlug}>
             Previous Event Names
-            <AnchorLink slug={urlify(`${eventName} Previous`)} />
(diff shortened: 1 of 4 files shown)


assets/js/analytics-events.tsx
```diff
@@ -102,18 +102,23 @@ function Event({ event }: { event: AnalyticsEvent }) {
...
+  const attributesSlug = urlify(`${eventName} Attributes`);
+  const exampleSlug = urlify(`${eventName} Example`);
+
   return (
     <div>
-      <h3>
+      <h3 id={slug}>
         {eventName}
-        <AnchorLink slug={urlify(eventName)} />
+        <AnchorLink slug={slug} />
```
zachmargolis, line 114: so having to repeat this change everywhere is maybe a hint that `AnchorLink` itself needs to be reworked.

We have `buildHeader()` that does an h-tag + anchor link together, maybe we just need more of these. However, that seems to have the same bug about not highlighting on first render that this PR has, so 🤷 

https://github.com/18F/identity-handbook/blob/38e461f2db6d53b96acc48b98466290cf65c52c4/assets/js/private-articles.tsx#L103-L110

aduth: Yeah I like the idea of making this into a component. I took a stab at it in 6569a7a (refactoring `buildHeader` in 1c2b72f).

zachmargolis: Love it! That didn't seem to fix the `:target` selector not working on first load, but maybe that can wait for another PR

aduth: > That didn't seem to fix the `:target` selector not working on first load, but maybe that can wait for another PR

Hm, that's interesting. I think it could be what you suspected as far as a race condition, like if the target styles would only be applied if it matches the hash fragment _at the time of initial page load_ (where JavaScript rendering would happen after).

aduth (approved):

PR #321, opened 2022-09-30
Upgrade ESBuild for JSX automatic runtime support

aduth:
**Why**: Simplifies implementation and helps align developer experience across projects (already available in IdP, for example).

Specifically, ESBuild v0.14.51 added support for the automatic JSX runtime ([see release notes](https://github.com/evanw/esbuild/releases/tag/v0.14.51)).


zachmargolis (approved):
LGTM

PR #322, opened 2022-09-30
Update Netlify CMS configuration for Cloud.gov Pages transition

aduth:
**Why**: [Federalist is becoming Cloud.gov Pages](https://cloud.gov/pages/federalist-migration/), and as such we'll need to update configurations accordingly.

I'm not sure of any resources for the specifics of these changes ([existing documentation](https://cloud.gov/pages/documentation/getting-started-with-netlify-cms/) still references Federalist), but these were the updates recommended directly from Cloud.gov Pages support folks.


zachmargolis (approved):
LGTM

PR #326, opened 2022-10-12
Reformat so that we can link directly to common task sections in gitlab docs

timothy-spencer:


zachmargolis (approved):
LGTM

If the goal is to change the levels so that the headings get anchors you can also set `toc_h_max: 4` or something so that h4's get anchors

timothy-spencer:
Oh, thank you!  I did not know that trick.  Well, this works too.

PR #332, opened 2022-10-21
Fix Netlify bug introduced by changing editorial workflow

jmdembe:
I am trying to fix a bug where if a user wants to create a new article, they are build fails and users are not able to submit new articles.

The new `netlify_sanbox.md` file _should_ make the test pass because there is a matching tile on the home page that will take you to the `netlify-sandbox` sandbox articles. I get false positives when testing locally, so I am going on what I read from the test and matching it to other categories.


admin/config.yml
```diff
@@ -18,6 +18,7 @@ collections:
     name: help
     folder: _articles
     create: true
+    path: '{{category}}/{{slug}}'
```
zachmargolis, line 21: is this the UI path? or the file path?
all articles are in the same directory (`_articles`) and `/articles`

zachmargolis (commented):
Not 100% sure I follow but that's because I am not an Netlify expert

admin/config.yml
jmdembe: yikes, that is the file path. I was messing with some things, but didn't remove this.

zachmargolis (approved):
LGTM

PR #337, opened 2022-10-25
Add links to source code from analytics events documentation

zachmargolis:
Related to: https://github.com/18F/identity-idp/pull/7208 

**Example:**

It adds links to [source definition](https://github.com/18f/identity-idp/blob/dfca5e2a/app/services/analytics_events.rb#L299), [source usages](https://github.com/search?q=email_language_updated+repo%3A18f%2Fidentity-idp+extension%3Arb&type=code)

**Screenshot:**

Screen Shot 2022-10-24 at 5 08 58 PM


assets/js/components/heading.tsx
```diff
@@ -15,11 +16,12 @@ export function Heading({
   level,
   children,
   id = urlify(children),
+  className,
 }: HeadingProps) {
```
aduth, line 20: Could also consider forwarding along any valid HTML props.

```suggestion
  ...props,
}: HeadingProps & JSX.HTMLAttributes<HTMLHeadingElement>) {
```

assets/js/github-urls.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return `https://github.com/${repo}/blob/${ref}/${file}${
+    line ? `#L${line}` : ""
+  }`;
+}
+
+export function searchURL({
+  needle,
+  repo,
+  extension,
+  path,
```
aduth, line 21: We don't ever pass this property. Should we? (e.g. `path: "app"`)

aduth (approved):

assets/js/github-urls.tsx
zachmargolis: I tried it out (specifically `app`) to exclude spec usages from search results, and I wasn't in love with it. But we could put it back if you think the results are better

assets/js/components/heading.tsx
zachmargolis: Ok! And we'd want to remove those from the definition of HeadingProps then, right?

aduth: Yeah there's a few more revisions that would need to be done in addition to that suggestion (also importing the `JSX` type and applying the props to `TagName`). The only thing we'd need to remove from the prop type is `className`.

assets/js/analytics-events.tsx
```diff
@@ -101,11 +121,44 @@ function Event({ event }: { event: AnalyticsEvent }) {
...
+            Source definition
+          </a>
+          ,{" "}
+          <a
+            className="usa-link"
+            href={searchURL({
+              needle: methodName,
+              repo: "18f/identity-idp",
+              extension: "rb",
+              type: "code",
```
zachmargolis, line 155:
```suggestion
              type: "code",
              path: "app",
```

assets/js/github-urls.tsx
zachmargolis: Updated in 265f764ea81e644666ad2db0420c5248456f4ba4

assets/js/components/heading.tsx
aduth: y'know, in retrospect, I think this is fine as-is, if you want to just leave it.

zachmargolis: LOL ok will leave it

PR #338, opened 2022-10-25
Add link to Device Profiling document

matthinz:
Add information about testing and configuration device identification / fraud detection.

### DIFF
--- a/_articles/device-identification.md
+++ b/_articles/device-identification.md
@@ -1,0 +1,7 @@
+---
+title: "Device identification and fraud detection"
+layout: article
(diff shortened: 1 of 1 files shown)


_articles/device-identification.md
```diff
@@ -0,0 +1,7 @@
+---
+title: "Device identification and fraud detection"
```
zachmargolis, line 2: I feel like we have been calling it "device profiling" everywhere, so that's the term I'd search for?
Otherwise LGTM

zachmargolis (approved):
LGTM

PR #340, opened 2022-10-26
Device identification -> device profiling

matthinz:
I had meant to include this commit on #338, but I think my firewall blocked my push. Anyway, this just switches around terminology to say "Device profiling" instead of "device identification"


aduth (approved):
LGTM

PR #341, opened 2022-10-26
Auto-convert times to local timezone for deploy schedules

zachmargolis:
Hypothesis: at the end of the day, all people really need to know for these schedules is UTC time, and their own time.

So we can simplify these giant tables with a billion columns into two columns

| before | after |
| ---- | ---- |
| Screen Shot 2022-10-26 at 3 49 47 PM | Screen Shot 2022-10-26 at 3 49 43 PM |

### DIFF
--- a/_articles/daily-deploy-schedule.md
+++ b/_articles/daily-deploy-schedule.md
@@ -26,10 +26,9 @@ In these environments
 The daily deploys are scheduled for **weekdays at 5pm UTC**. See
 the [daily business schedule in identity-devops][identity-devops-schedule].
 
-| UTC  | Daylight Savings               | Eastern | Central | Mountain | Pacific |
-|------|--------------------------------------|------|------|------|------|------|
-| 5pm  | Daylight savings time<br />(March-November) | 1pm  | 12pm | 11am | 10am |
-|      | Standard time                               | 12pm | 11am | 10am | 9am  |
+| UTC  | Local (<span data-local-tzname></span>) |
+|------|-----------------------------------------|
+| 5pm  | <span data-utc-time="5pm"></span>       |
 
 [identity-devops-schedule]: https://github.com/18F/identity-terraform/blob/3a37047cfae6949dab1150025c528ccc5332f837/asg_recycle/main.tf#L44-L78
 
@@ -42,13 +41,9 @@ In these environments
 
 Deploys are scheduled every 6 hours at **5am, 11am, 5pm, 11pm UTC**.
 
-| UTC  | Daylight Savings               | Eastern | Central | Mountain | Pacific |
-|------|--------------------------------------|------|------|------|------|------|
-| 5am  | Daylight savings time<br />(March-November) | 1am  | 12am | 11pm | 10pm |
-|      | Standard time                               | 12am | 11pm | 10pm | 9pm  |
-| 11am | Daylight savings time<br />(March-November) | 7am  | 6am  | 5am  | 4am  |
-|      | Standard time                               | 6am  | 5am  | 4am  | 3am  |
-| 5pm  | Daylight savings time<br />(March-November) | 1pm  | 12pm | 11am | 10am |
-|      | Standard time                               | 12pm | 11am | 10am | 9am  |
-| 11pm | Daylight savings time<br />(March-November) | 7pm  | 6pm  | 5pm  | 4pm  |
-|      | Standard time                               | 6pm  | 5pm  | 4pm  | 3pm  |
+| UTC  | Local (<span data-local-tzname></span>) |
+|------|-----------------------------------------|
+| 5am  | <span data-utc-time="5pm"></span>       |
+| 11am | <span data-utc-time="11am"></span>      |
+| 5pm  | <span data-utc-time="5pm"></span>       |
+| 11pm | <span data-utc-time="11pm"></span>      |
--- a/_articles/daily-deploy-schedule.md
+++ b/_articles/daily-deploy-schedule.md
@@ -26,9 +26,9 @@ In these environments
 The daily deploys are scheduled for **weekdays at 5pm UTC**. See
 the [daily business schedule in identity-devops][identity-devops-schedule].
 
-| UTC  | Local (<span data-local-tzname></span>) |
-|------|-----------------------------------------|
-| 5pm  | <span data-utc-time="5pm"></span>       |
+| UTC  | Local (<lg-local-zone-name />) |
+|------|--------------------------------|
+| 5pm  | <lg-local-time utc="5pm" />    |
 
 [identity-devops-schedule]: https://github.com/18F/identity-terraform/blob/3a37047cfae6949dab1150025c528ccc5332f837/asg_recycle/main.tf#L44-L78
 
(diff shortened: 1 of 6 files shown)


_articles/daily-deploy-schedule.md
```diff
@@ -26,10 +26,9 @@ In these environments
 The daily deploys are scheduled for **weekdays at 5pm UTC**. See
 the [daily business schedule in identity-devops][identity-devops-schedule].
 
-| UTC  | Daylight Savings               | Eastern | Central | Mountain | Pacific |
-|------|--------------------------------------|------|------|------|------|------|
-| 5pm  | Daylight savings time<br />(March-November) | 1pm  | 12pm | 11am | 10am |
-|      | Standard time                               | 12pm | 11am | 10am | 9am  |
+| UTC  | Local (<span data-local-tzname></span>) |
+|------|-----------------------------------------|
+| 5pm  | <span data-utc-time="5pm"></span>       |
```
aduth, line 31: Could be a good candidate for a custom element, which would avoid the querying / "set up" logic.

assets/js/setup.tsx
```diff
@@ -60,4 +61,21 @@ export const loadSimpleJekyllSearch = () => {
   });
 };
 
+export const setUpTimezoneHooks = () => {
+  Array.from(
+    document.querySelectorAll<HTMLElement>("[data-local-tzname]")
+  ).forEach((elem) => {
+    elem.innerText = localTimezoneName();
+  });
```
aduth, line 69: You could use [`textContent`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) to avoid specifying the generic. (Same applies below)

```suggestion
  Array.from(document.querySelectorAll("[data-local-tzname]")).forEach(
    (elem) => {
      elem.textContent = localTimezoneName();
    }
  );
```

assets/js/timezone.tsx
```diff
@@ -0,0 +1,36 @@
+export function localTimezoneName() {
+  return Intl.DateTimeFormat().resolvedOptions().timeZone;
+}
+
+/**
+ * @param utcTime string of time in UTC such as "5pm", "6am", "6:30am"
+ * @return converted to local time such as "8pm", "2:30pm"
+ */
+export function convertToLocal(utcTime: string) {
```
aduth, line 9: Neat function! 

Alternatively, another option which is more compact at the cost of some flexibility and usability, we could initialize it with a UTC hour and use `Intl.DateTimeFormat` ?

```ts
new Intl.DateTimeFormat(undefined, { hour: 'numeric' }).format(new Date().setUTCHours(hours));
```

assets/js/timezone.tsx
```diff
@@ -0,0 +1,36 @@
+export function localTimezoneName() {
```
aduth, line 1: There's no JSX in this file so I think the file extension should be `.ts` to avoid confusion.

assets/js/timezone.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * @return converted to local time such as "8pm", "2:30pm"
+ */
+export function convertToLocal(utcTime: string) {
+  const match = utcTime.match(/(\d+)(:(\d+))?\s*([ap].?m)/i);
+
+  if (!match) {
+    return "N/A";
+  }
+
+  const [, hourStr, , minuteStr, ampm] = match;
```
aduth, line 16: Do we need the minutes support?

aduth (approved):
Nice improvement!

assets/js/timezone.tsx
zachmargolis: For most of us in the US, no. But what if somebody [from a :30 or :45 Australia timezone](https://www.timeanddate.com/time/australia/time-zones-background.html) views the page?

aduth: > For most of us in the US, no. But what if somebody [from a :30 or :45 Australia timezone](https://www.timeanddate.com/time/australia/time-zones-background.html) views the page?

😮 

There must be a "falsehoods programmers believe about X" article regarding times that I missed 😅

jmdembe: > > For most of us in the US, no. But what if somebody [from a :30 or :45 Australia timezone](https://www.timeanddate.com/time/australia/time-zones-background.html) views the page?
> 
> 😮
> 
> There must be a "falsehoods programmers believe about X" article regarding times that I missed 😅

I have just the article for that: [falsehoods programmers believe about timezones](https://www.zainrizvi.io/blog/falsehoods-programmers-believe-about-time-zones/)

assets/js/timezone.tsx
zachmargolis: gave that a shot as part of other changes in a46c3555e65a46619e7d56b858b1ed87dbfac7da

assets/js/timezone.tsx
zachmargolis: updated in a46c3555e65a46619e7d56b858b1ed87dbfac7da

_articles/daily-deploy-schedule.md
zachmargolis: I present `<lg-local-zone-name />` and `<lg-local-time utc="5pm" />` in a46c3555e65a46619e7d56b858b1ed87dbfac7da

assets/js/setup.tsx
zachmargolis: went with customelements approach in a46c3555e65a46619e7d56b858b1ed87dbfac7da

assets/js/timezone.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  // Comes back like "7:01 PM"
+  const formatted = new Intl.DateTimeFormat(undefined, {
+    hour: "numeric",
+    minute: "2-digit",
+    hour12: true,
+  }).format(localDate);
+
+  return formatted.toLowerCase().replace(" ", "").replace(":00", "");
```
aduth, line 35: Wondering about the first argument (locale) of `DateTimeFormat` and the `hour12` option:

- By specifying `undefined`, we're relying on the user's locale, and it doesn't seem like we should hard-control the 12-hour setting, and our string replacements may not be totally accurate.
- Alternatively, we could force it to `'en-US'` which makes it more predictable (and probably still don't need `hour12`), and would be consistent with the rest of the page otherwise being in English

```suggestion
  const formatted = new Intl.DateTimeFormat('en-US', {
    hour: "numeric",
    minute: "2-digit",
  }).format(localDate);

  return formatted.toLowerCase().replace(" ", "").replace(":00", "");
```

assets/js/timezone.ts
```diff
@@ -0,0 +1,57 @@
...
+    const utcTime = this.getAttribute("utc");
+    if (utcTime) {
+      this.textContent = convertToLocal(utcTime);
+    }
+  }
+}
+
+export function installCustomTimeElements() {
+  customElements.define("lg-local-zone-name", LocalZoneName);
+  customElements.define("lg-local-time", LocalTime);
```
aduth, line 55: As a convention, I might suggest suffixing the element classes with "Element"

```suggestion
class LocalZoneNameElement extends HTMLElement {
  connectedCallback() {
    this.textContent = localTimezoneName();
  }
}

class LocalTimeElement extends HTMLElement {
  connectedCallback() {
    const utcTime = this.getAttribute("utc");
    if (utcTime) {
      this.textContent = convertToLocal(utcTime);
    }
  }
}

export function installCustomTimeElements() {
  customElements.define("lg-local-zone-name", LocalZoneNameElement);
  customElements.define("lg-local-time", LocalTimeElement);
```

assets/js/timezone.tsx
aduth: > > > For most of us in the US, no. But what if somebody [from a :30 or :45 Australia timezone](https://www.timeanddate.com/time/australia/time-zones-background.html) views the page?
> > 
> > 
> > 😮
> > There must be a "falsehoods programmers believe about X" article regarding times that I missed 😅
> 
> I have just the article for that: [falsehoods programmers believe about timezones](https://www.zainrizvi.io/blog/falsehoods-programmers-believe-about-time-zones/)

Thanks for sharing!

PR #352, opened 2022-12-05
Skip skipping migrations so we do not trip

pauldoomgov:
There are a lot of "ifs" in using --skip-migrations which means lots of ways to get cut.   I suggest we pull it from the handbook for now.


zachmargolis (approved):
LGTM! Maybe we ditch the whole "config recycle" section since it's now just a regular recycle?

zachmargolis:
Gonna merge this and we can always make another PR to update again if needed!

PR #354, opened 2022-12-14
Add information about time before monitoring migrations

jmdembe:
This PR updates the amount of time to wait when starting to monitor migrations

### DIFF
--- a/_articles/appdev-deploy.md
+++ b/_articles/appdev-deploy.md
@@ -147,7 +147,7 @@ Staging used to be deployed by this process, but this was changed to deploy the 
 
    1. Follow the progress of the migrations, ensure that they are working properly <a id="follow-the-process" />
    ```bash
-   # may need to wait a few seconds after the recycle
+   # may need to wait ~5 minutes after the recycle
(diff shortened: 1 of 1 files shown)


_articles/appdev-deploy.md
```diff
@@ -147,7 +147,7 @@ Staging used to be deployed by this process, but this was changed to deploy the
 
    1. Follow the progress of the migrations, ensure that they are working properly <a id="follow-the-process" />
    ```bash
-   # may need to wait a few seconds after the recycle
+   # may need to wait ~5 minutes after the recycle
```
aduth, line 150: Does it often take this long? I agree "a few seconds" is probably shorter than realistic. Maybe we can split the difference?

```suggestion
   # may need to wait a few minutes after the recycle
```

aduth (approved):
LGTM!

_articles/appdev-deploy.md
jmdembe: That works. I gave a rough estimate here.

PR #356, opened 2022-12-30
Service provider linked accounts [LG-8518]

gitmijn:
This is a routine query we do for some service providers to provide a count of total users for their application.

### DIFF
--- a/_articles/queries.md
+++ b/_articles/queries.md
@@ -52,12 +52,12 @@ RegistrationLog.where('registered_at <= ?', date).count
 Returns the total number of registered users linked to a service provider or service providers.
 
 ```ruby
-
-User.includes( :service_providers ).where({ :service_providers => {:issuer => ['urn:gov:gsa:openidconnect:test', 'urn:gov:gsa:openidconnect:test_prompt_login_banned' ]}}).count
+User.joins(:service_providers ).where(
+  service_providers: {
+    issuer: ['issuer:a', 'issuer:b']
+  }
+).count
 ```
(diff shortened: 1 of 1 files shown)


_articles/queries.md
```diff
@@ -30,6 +30,8 @@ date = Date.new(2021, 1, 1)
 User.where('created_at <= ?', date).count
 ```
 
+
+
```
zachmargolis, line 34:
```suggestion
```
no need for these extra lines IMO

_articles/queries.md
```diff
@@ -47,6 +49,18 @@ date = Date.new(2021, 1, 1)
...
+Returns the total number of registered users linked to a service provider or service providers.
+
+```ruby
+
+User.includes( :service_providers ).where({ :service_providers => {:issuer => ['urn:gov:gsa:openidconnect:test', 'urn:gov:gsa:openidconnect:test_prompt_login_banned' ]}}).count
+```
+
+**Note**: a `.joins` clause would be a good option here instead of a `.includes` 
+
```
zachmargolis, line 62: I'd just make the example use a `.joins` instead of giving the option. Also switching to new-style hash syntax and adding some different line breaks:

~~~suggestion
```ruby
User.joins(:service_providers ).where(
  service_providers: {
    issuer: ['issuer:a', 'issuer:b']
  }
).count
```
~~~

zachmargolis (approved):
thanks for adding this so quickly! small suggestions but I love it

PR #360, opened 2023-01-04
Update analytics-events page to be handle nullable "types" field

zachmargolis:
Workaround to handle bug fixed by https://github.com/18F/identity-idp/pull/7575


aduth (approved):
LGTM 👍

PR #364, opened 2023-01-11
Add oncall instructions for password reset

zachmargolis:
This is temporary until we can push partners to use our RISC API to self-report these

### DIFF
--- a/_articles/appdev-oncall-guide.md
+++ b/_articles/appdev-oncall-guide.md
@@ -73,7 +73,7 @@ Throughout the week, check for automated vulnerability pull requests and try to 
 
 ### Resetting User Passwords
 
-On rare occasions partners will ask us to reset passwords for accounts. In a Rails console, run:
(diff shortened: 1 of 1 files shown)


_articles/appdev-oncall-guide.md
```diff
@@ -71,6 +71,23 @@ Throughout the week, check for automated vulnerability pull requests and try to
 
 [ig-requests]: https://github.com/18F/identity-security-private/wiki/Responding-to-Inspector-General-(IG)-Data-Requests
 
+### Resetting User Passwords
+
+On rare occasions partners will ask us to reset passwords for accounts. In a Rails console, run:
```
orenyk, line 76: **question:** does this need to be a Rails console with write DB access?

zachmargolis: Yes, good point. Updated in 1b06da2841c79c3b413d24d182548b32943cd343

PR #375, opened 2023-02-03
Update queries to include notes about accounts that were deleted

zachmargolis:
Based on work I did in https://github.com/18F/identity-idp/pull/7731

### DIFF
--- a/_articles/queries.md
+++ b/_articles/queries.md
@@ -17,7 +17,7 @@ ActiveRecord::Base.connection.execute 'SET statement_timeout = 200000'
 
 ## Total Registered
 
-Returns the number of users that created accounts (this includes users who may not have fully registered, see [Fully Registered Users](#fully-registered-users) for that query)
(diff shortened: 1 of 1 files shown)


_articles/queries.md
```diff
@@ -17,7 +17,7 @@ ActiveRecord::Base.connection.execute 'SET statement_timeout = 200000'
 
 ## Total Registered
 
-Returns the number of users that created accounts (this includes users who may not have fully registered, see [Fully Registered Users](#fully-registered-users) for that query)
+Returns the number of users that created accounts (this includes users who may not have fully registered, see [Fully Registered Users](#fully-registered-users) for that query), it does not count users who have already deleted their accounts.
```
jmdembe, line 20: Very, very small suggestion, but other than that, it 👍🏾 

```suggestion
Returns the number of users that created accounts (this includes users who may not have fully registered, see [Fully Registered Users](#fully-registered-users) for that query). The count does not include users who have already deleted their accounts.
```

zachmargolis: Love it, thanks!

jmdembe (approved):

PR #379, opened 2023-02-16
WIP - Change Comms Lead role to Messenger and note IR comms threshold criteria

pauldoomgov:
Based on recent tabletop exercises and refinement of both IR and crisis communications process, this deconflicts the "Comms Lead" role and changes it to something more accurate: "Messenger"

This should be a low risk change for the informal IR role names and prevent confusion.


zachmargolis (approved):
LGTM

jeffthink (approved):
🚢

PR #380, opened 2023-02-17
Mark sprint teams that are not active as archived

zachmargolis:
Grays out teams and adds a lil note:

| example |
| --- |
| Screen Shot 2023-02-17 at 11 20 04 AM |


pauldoomgov (approved):
LGTM

aduth (approved):
LGTM 👍 

Maybe for a follow-on, but should we consider sorting them in a separate grouping from the active teams?

zachmargolis:
> Maybe for a follow-on, but should we consider sorting them in a separate grouping from the active teams?

I'm open to that! But don't have the energy for it right now

PR #385, opened 2023-03-08
Add spellchecking via cspell

zachmargolis:
Copied liberally from https://github.com/18F/handbook/pull/3491


_articles/appdev-troubleshooting-production.md
```diff
@@ -117,7 +117,7 @@ We don't have indexed lookups by phone number so we need to combine a few approa
     ```
 
 2. In Cloudwatch, look up OTP requests to those fingerprints. It helps to have an approximate date
-   this user was active to help narrow down the Cloudwatch search timeframe.
+   this user was active to help narrow down the Cloudwatch search time frame.
```
aduth, line 120: TIL that _time frame_ is two words!

aduth (approved):
This is really neat! It found a lot of issues, which is equally good to get sorted and also disappointing to have lived in our handbook for so long 😅 

Also, that's quite a big dependency tree for the new package, but I suppose it's doing quite a bit of work to do the analysis 🤷

_articles/appdev-troubleshooting-production.md
zachmargolis: Human languages are moving targets! Nothing is set in stone! I feel like we could make a statement that it's one word in our usage if we cared. There are reference to both one- and two-word forms of it

zachmargolis: Leaving as two words for now but we can always re-evaluate later

aduth: Yeah, no strong stance from me on this one, just surprising.

SeanKilleen:
Just wanted to say it's great to see my original PR spread further! 🎉 Just happened to see this.

PR #390, opened 2023-03-16
Set Ruby version to 2.7.5

aduth:
Why: Fix failing builds due to Cloud.gov Pages Ruby compilation errors.

See: https://gsa-tts.slack.com/archives/C04823NMA/p1678797846444569


zachmargolis (approved):
LGTM

PR #392, opened 2023-03-16
Update deployment guide to adjust cadence

aduth:
Why: As decided in Engineering Huddle 2023-03-15, we'll now be deploying on Tuesdays and Thursdays instead of Mondays and Thursdays.


zachmargolis (approved):
LGTM

PR #393, opened 2023-03-16
Rename expiring cert guide to fix typo

aduth:
Why: Spelling is important


pauldoomgov (approved):
I prefer to toubleshoot, but to each their own...

LGTM!

PR #396, opened 2023-04-03
Add link to SBAR template from handbook

zachmargolis:
To help me (and maybe others) find this in the future

### DIFF
--- a/_articles/glossary.md
+++ b/_articles/glossary.md
@@ -254,6 +254,11 @@ A trusted entity that establishes and vouches for the identity or attributes of 
 ### SAML
 **Security Assertion Markup Language**
 
+### SBAR
+**Situation, Background, Assessment, Recommendation**
+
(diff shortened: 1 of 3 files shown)


benjaminchait (approved):
@zachmargolis - This seems fine. Should we also include a "copy me" link? (I don't know how those are generated, you might need to help teach me!)

zachmargolis:
> @zachmargolis - This seems fine. Should we also include a "copy me" link? (I don't know how those are generated, you might need to help teach me!)

1. to make the "copy me" link, you take the `/edit` at the end of the URL and switch to `/copy`
2. for this one, it's just a plain redirect from the articles list. I feel like I'd like to have a view link and a copy link, so would you prefer if I stubbed out an article explaining SBARs a bit that had both links? (similar to the [rollplans](https://handbook.login.gov/articles/rollplans.html) page?)

Screen Shot 2023-04-03 at 8 54 24 AM

benjaminchait:
@zachmargolis - Yeah, I like the idea of an article that has both links. Do you have bandwidth to cover that? (If not, I can take a swing at it later this week...)

PR #398, opened 2023-04-04
Build JavaScript files to site output directory

aduth:
This is essentially the same as what had been implemented in the brochure site in https://github.com/18F/identity-site/pull/1011, and all of the same benefits described there apply here. This should also fix an error which sometimes occurs due to Jekyll watching JavaScript assets which are deleted as part of the JavaScript build ("No such file or directory @ rb_file_s_stat").

Testing instructions:

- Ensure preview JavaScript behaviors work as expected (no 404s in network tab, search and anchor links initialize as expected)
- Ensure that local Jekyll build does not clobber JavaScript assets:

```
rm -rf _site
npm run build
bundle exec jekyll build
ls _site/assets/js
```

### DIFF
--- a/_articles/analytics-events.md
+++ b/_articles/analytics-events.md
@@ -80,7 +80,7 @@ For additional events see the [legacy Analytics Event documentation][legacy-docu
 </div>
 
 <script type="module">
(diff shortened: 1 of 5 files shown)


zachmargolis (approved):
makes it look easy! thanks!

PR #400, opened 2023-04-05
Define build dependencies as dependencies

aduth:
The moves any dependencies required for build from `devDependencies` to `dependencies`, to support the assignment of `NODE_ENV=production` environment variable in build environments, since `NODE_ENV=production` will cause `npm install` to not install `devDependencies`. `NODE_ENV=production` will trigger build-time optimizations such as CSS minification (via `@18f/identity-build-sass`) which are expected for the live sites.

This is the same idea as https://github.com/18F/identity-site/pull/1065

This also updates the moved dependency `esbuild` to the latest version (from 0.15.10 to 0.17.15).

**Testing Instructions:**

```
rm -rf _site node_modules
export NODE_ENV=production
npm install
make build
```

Previously:

```
[build:js:prod] sh: esbuild: command not found
[build:js:prod] npm run build:js:prod exited with code 127
```


zachmargolis (approved):
LGTM

PR #401, opened 2023-04-05
Note new idv_unavailable flag in vendor outage response process

matthinz:
Issue: [LG-8710](https://cm-jira.usa.gov/browse/LG-8710)

We've added a new config flag, `idv_available`, that can be used to disable remote unsupervised identity verification (this augments the existing `vendor_status_*` flags).

This PR updates the vendor outage response process documentation to include a note about this flag. 

(I since I was in there, I also removed references to `partial_outage`--support for this was not implemented.)


zachmargolis (approved):
LGTM

PR #404, opened 2023-04-06
Add redirect to New Feature Onboarding checklist

zachmargolis:


_articles/new-feature-onboarding.md
```diff
@@ -0,0 +1,7 @@
+---
+title: "New Feature Onboarding"
+description: "Checklist for new features, specifically focused on documentation needed for security review"
+layout: article
+category: "Product"
```
aduth, line 5: I would _love_ to have something like this for general product features, but this document seems very tailored to security reviews. Maybe it should be categorized accordingly?

aduth (approved):

_articles/new-feature-onboarding.md
zachmargolis: Are you suggesting this?
```suggestion
category: "Security"
```

aduth: Yes 👍

aduth: I think there's a lot of overlap with a hypothetical product feature checklist though, so I don't feel _too_ strongly about it 🤷

zachmargolis: When that checklist exist, we can update + recombine!

PR #405, opened 2023-04-06
update gitlab docs to use 2023 saml

timothy-spencer:
Also add a missing line so that the cert will be proper.

### DIFF
--- a/_articles/platform-gitlab.md
+++ b/_articles/platform-gitlab.md
@@ -121,14 +121,15 @@ GitLab leverages Omniauth to allow users to sign in using a variety of services,
 openssl req -nodes -x509 -days 365 -newkey rsa:2048 -keyout private.pem -out public.crt
 ```
 
-1. Grab the IDP sandbox signing certificate from <https://developers.login.gov/saml/> and get its fingerprint:
+1. Grab the IDP sandbox signing certificate from <https://developers.login.gov/saml/> and get its fingerprint (use secure.login.gov instead of idp.int.identitysandbox.gov if you are doing this for gitlab prod/staging):
 ```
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM

_articles/platform-gitlab.md
```diff
@@ -121,14 +121,15 @@ GitLab leverages Omniauth to allow users to sign in using a variety of services,
 openssl req -nodes -x509 -days 365 -newkey rsa:2048 -keyout private.pem -out public.crt
 ```
 
-1. Grab the IDP sandbox signing certificate from <https://developers.login.gov/saml/> and get its fingerprint:
+1. Grab the IDP sandbox signing certificate from <https://developers.login.gov/saml/> and get its fingerprint (use secure.login.gov instead of idp.int.identitysandbox.gov if you are doing this for gitlab prod/staging):
 ```
-curl -s https://idp.int.identitysandbox.gov/api/saml/metadata2021 \
+curl -s https://idp.int.identitysandbox.gov/api/saml/metadata2023 \
```
zachmargolis, line 126: if we wanted, we could use jekyll to magically always use the current year?
```suggestion
curl -s https://idp.int.identitysandbox.gov/api/saml/metadata{{ site.time | date: '%Y' }}
 \
```

timothy-spencer: maaaaybe?  I've never done that.  Is there ever any time in 202X where the 202X saml cert won't be available?

zachmargolis: Ideally our process would be smooth like that, but I'm not sure how smooth it actually is

PR #416, opened 2023-04-13
Fix broken markdown link

zachmargolis:
It was amateur hour for me this week, clearly

| before | after |
| ---- | --- |
|  Screen Shot 2023-04-13 at 4 07 24 PM | Screen Shot 2023-04-13 at 4 07 19 PM |

### DIFF
--- a/_articles/eng-design-doc.md
+++ b/_articles/eng-design-doc.md
@@ -3,6 +3,6 @@ title: "Engineering Design Doc Template"
 layout: article
 category: "AppDev"
 redirect_to: https://docs.google.com/document/d/1KlQL4MMCXahV12mMbh5scmXLl6adjv_0fcFzQ8si7qc/edit#
-description: "Template that can be copied for engineering proposals, replaced in favor of [RFCs](/articles/product-artifacts.html#rfcs"
+description: "Template that can be copied for engineering proposals, replaced in favor of [RFCs](/articles/product-artifacts.html#rfcs)"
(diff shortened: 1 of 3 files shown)


_articles/eng-design-doc.md
```diff
@@ -3,6 +3,6 @@ title: "Engineering Design Doc Template"
 layout: article
 category: "AppDev"
 redirect_to: https://docs.google.com/document/d/1KlQL4MMCXahV12mMbh5scmXLl6adjv_0fcFzQ8si7qc/edit#
-description: "Template that can be copied for engineering proposals, replaced in favor of [RFCs](/articles/product-artifacts.html#rfcs"
+description: "Template that can be copied for engineering proposals, replaced in favor of [RFCs](/articles/product-artifacts.html#rfcs)"
```
aduth, line 6: Might be made difficult by the fact it's in the YAML frontmatter, but is there a way we can make this link work in the preview environments?

aduth (approved):
LGTM

_articles/eng-design-doc.md
zachmargolis: I think that's why identity-site had $BASE_URL manual substitution:

https://github.com/18F/identity-site/blob/aa74380cc5701b9824008ee9ef54e865a201f17c/_includes/components/footer-navigation.html#L34

so we could add it to the template, but I have one more idea to maybe redo this

PR #422, opened 2023-04-24
Update language about 360s

micahsaul:
Removing line about contractors to align with conversations at the leadership level.


zachmargolis (approved):
LGTM

PR #426, opened 2023-05-02
Upgrade Login.gov Design System to v7.0.1

aduth:
Upgrades the Login.gov Design System from v6.5.0 to v7.0.1 (currently beta, but will target stable release prior to merge).

Preview link: https://federalist-7d0b2b76-42fc-4df1-9825-8c3cd77e0a15.sites.pages.cloud.gov/preview/18f/identity-handbook/aduth-lgds-7/
Compare to live: https://handbook.login.gov/

One difference/regression I notice here that might be worth resolving before the stable release of 7.0.1 is that the handbook's use of the "basic" header lost the vertical centering of the right-hand content in the refactoring of https://github.com/18F/identity-design-system/pull/352 .

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -15,39 +15,29 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   </head>
   <body data-base-url="{{site.baseurl}}">
-    <header class="usa-header usa-header--basic" role="banner">
-      <div class="usa-nav-container">
-        <div class="usa-navbar">
-          <div class="usa-logo" id="basic-logo">
-            <a href="/" title="Home" aria-label="Home">
-              <img src="{{site.baseurl}}/assets/img/login-gov-logo.svg" class="usa-logo__img" alt="Login.gov" />
-              <em class="usa-logo__text">Handbook</em>
-            </a>
-          </div>
-          <!--
-            <button class="usa-menu-btn">Menu</button>
-          -->
+    <header class="usa-header usa-header--extended">
+      <div class="usa-navbar">
+        <div class="usa-logo">
+          <a href="/" title="Home" aria-label="Home">
+            <img src="{{ site.baseurl }}/assets/img/login-gov-logo.svg" class="usa-logo__img" alt="Login.gov" />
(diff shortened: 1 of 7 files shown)


zachmargolis:
> One difference/regression I notice here that might be worth resolving before the stable release of 7.0.1 is that the handbook's use of the "basic" header lost the vertical centering of the right-hand content in the refactoring of [18F/identity-design-system#352](https://github.com/18F/identity-design-system/pull/352) .

seems like it made the header a bit taller? I'm not that worried about it

aduth:
I pushed a refactor to the header markup in 3218668 which should have it display a bit nicer now.

_layouts/default.html
```diff
@@ -15,39 +15,29 @@
...
+          <div class="flex-row flex-align-center grid-row grid-gap-1">
+            <ul class="usa-nav__secondary-links grid-col grid-col-auto margin-0">
+              <li class="usa-nav__secondary-item" id="private-login-button-container"></li>
+            </ul>
+            <form class="usa-search usa-search--small grid-col grid-col-fill" role="search">
+              <label class="usa-sr-only" for="search-input">Search articles</label>
+              <input class="usa-input width-auto" type="search" id="search-input" placeholder="Search articles...">
+              <button class="usa-button" type="submit">
+                <span class="usa-search__submit-text">Search</span>
+              </button>
```
zachmargolis, line 36: Looks like the search box changed, we didn't used to have a submit button, and since the search is all type-as-you go, maybe we don't need it?

| before | after |
|----|----|
| Screen Shot 2023-05-03 at 9 05 20 AM | Screen Shot 2023-05-03 at 9 05 12 AM |

maybe we just remove that?
```suggestion
```

aduth: That was intentional, partly because the search input shows strangely without the adjacent button (without a right border, etc, since it expects the button to be there). Also partly as a small accessibility/usability improvement, to allow it to operate as a submittable form and programmatically shift focus into the results when submitted.

zachmargolis: sounds good! I noticed all the JS/on-submit changes too after I posted that 🙈

PR #427, opened 2023-05-10
Added partner feedback/kudos g-sheets

rahul-kamarouthu:
Previously the page had a google doc as a placeholder and is now being swapped with a google sheet with partner feedback/kudos.

Waiting for confirmation from AM team.


zachmargolis (approved):
LGTM

PR #430, opened 2023-05-15
use ALL option with scale-remove-*-instances scripts (vs. separate idp / worker calls)

bleachbyte:
Does what it says on the tin!


zachmargolis (approved):
LGTM

PR #434, opened 2023-05-24
Add glossary terms related to in-person proofing with USPS

NavaTim:
References:
- https://about.usps.com/postal-bulletin/2020/pb22556/html/cover_004.htm
- https://www.uspsoig.gov/reports/audit-reports/retail-systems-software-deployment-and-functionality

### DIFF
--- a/_articles/glossary.md
+++ b/_articles/glossary.md
@@ -194,6 +194,11 @@ To differentiate, we could say:
 
 > The user proofed their identity to complete identity verification.
 
+### IVS
+**Identity Verification Services**
+
+This refers to part of the Identity and Access Management portfolio of the U.S. Postal Service.
+
 ### LOA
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM! What if we linked the reference URLs in the doc itself?

NavaTim:
> LGTM! What if we linked the reference URLs in the doc itself?

@zachmargolis Wasn't sure about doing that since they're dated publications instead of regular static pages.

zachmargolis:
> > LGTM! What if we linked the reference URLs in the doc itself?
> 
> @zachmargolis Wasn't sure about doing that since they're dated publications instead of regular static pages.

To me, some context seems better than none, but defining them at all is still good

mitchellhenke (approved):

PR #435, opened 2023-06-05
Update offboarding.md

benjaminchait:
We have a team member who is rolling-off Team Ada later this month, and I was asked to ensure this [offboarding](https://handbook.login.gov/articles/offboarding.html) was completed. While reviewing that checklist, I noticed a few things were out-of-date, so this is a Pull Request to suggest edits. (I am open to us choosing to do less - or more! - with this change.)

In no particular order...
1. I don't think the referenced People-Ops JIRA board is still used (it shows either as hidden or deleted)
2. The TTS Slack Form is now in ServiceNow; but it's unclear if we need that form, or if it should be under "Services"
3. The "all accounts" section links to [Services and Accounts](https://handbook.login.gov/articles/accounts.html) which seems better than duplicating content here

### DIFF
--- a/_articles/offboarding.md
+++ b/_articles/offboarding.md
@@ -15,25 +15,12 @@ Review the [Leaving TTS page in the TTS Handbook](https://handbook.tts.gsa.gov/l
 
 ### For offboarding assistant to complete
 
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM

PR #437, opened 2023-06-09
Update ruby to 3.2.2, Jekyll to 4.3.2

solipet:
I was unable to build `event_machine` under ruby 2.7.5 on my mac. After updating ruby to 3.2.2, Jekyll fails with a call to `String#tainted?` which was removed in 3.?. Updating Jekyll resolved this.

**Note:** I have no idea how this update affect the deployed handbook - need to confirm this is safe before merging.


aduth (approved):
LGTM 👍 

I think this should be safe to upgrade now. Previously we were waiting on Cloud.gov Pages support, but between their [recent blog post](https://cloud.gov/2023/05/24/cloud-gov-pages-jekyll-ruby-upgrade/) and the fact the preview site looks fine, I think it should be okay.

PR #441, opened 2023-06-20
Jmax/lg 8870 update vendor outage runbook

jmax-gsa:
Updated section on Identity Verification outage flags.


zachmargolis (approved):
LGTM, I would also add @18F/identity-ada as reviewers too? Since that's the team with more context on these changes?

_articles/vendor-outage-response-process.md
```diff
@@ -60,26 +60,88 @@ The possible values for each flag:
...
+
+ - The `SMS` checkbox on the IdV dialogs will be disabled.
+ 
+ - New phone numbers for SMS 2FA will not be verified.
+
+  - Any user entering the verification flow will be presented with an
+    outage screen telling them that their options are to verify by
+    mail or wait until our vendor outage is resolved.
+
+  - Users will only be able to upload their IDs from their computer;
```
benjaminchait, line 134: Minor nit: Users who _start_ on a mobile phone will be okay, but the _hybrid handoff_ will not be available.

What do you think of: "Users will need to add photos of their ID from the device on which they start; desktop users will not be presented the opportunity to switch to a mobile phone for adding their ID."

_articles/vendor-outage-response-process.md
```diff
@@ -60,26 +60,88 @@ The possible values for each flag:
...
+#### Voice
+  - Users will only be able to verify their identity by mail.
+  
+  - New phone numbers for voice 2FA will not be verified
+ 
+  - Any user entering the verification flow will be presented with an
+    outage screen telling them that their options are to verify by
+    mail or wait until our vendor outage is resolved.
+
+  - Users will only be able to upload their IDs from their computer;
```
benjaminchait, line 146: I don't think we use the Pinpoint "voice" feature for hybrid handoff, so I think this line could be removed? 

But all other content in this section makes sense!

benjaminchait (changes requested):

_articles/vendor-outage-response-process.md
jmax-gsa: Re-worded your suggestion a bit; see if you like it.

_articles/vendor-outage-response-process.md
jmax-gsa: Nope. That's how the system behaves right now. I'm on board with changing it; I'll even write the ticket if you like. But right now, we don't offer hybrid handoff if voice is disabled.

The spec at `features/idv/outage_spec:185` check for exactly this behavior, and I just manually verified it to be extra doubleplus sure.

benjaminchait: Ah, that is...surprising!

Yes, can you drop a ticket on Team Ada's board, and we can refine/update things accordingly?

PR #447, opened 2023-07-05
Update devops-scripts.md

Sgtpluck:
Update directions for deploying configuration so it doesn't reference passenger restart, which is not safe to use,


zachmargolis (approved):
LGTM thx for fixing!

PR #455, opened 2023-07-26
Suggest using ls-servers before deploying

soniaconnolly:
It's a good idea to check if there are extra instances running or if there are issues with the current instances before kicking off a deploy.

### DIFF
--- a/_articles/appdev-deploy.md
+++ b/_articles/appdev-deploy.md
@@ -142,7 +142,11 @@ Staging used to be deployed by this process, but this was changed to deploy the 
    ```bash
    cd identity-devops
    ```
-1. Recycle the IDP instances to get the new code, it automatically creates a new migration instance first.
+1. Check current server status, and confirm that there aren't extra servers running. If there are, scale in old instances before deploying.
+   ```bash
+   aws-vault exec prod-power -- ./bin/ls-servers -e prod
+   ```
+1. Recycle the IDP instances to get the new code. It automatically creates a new migration instance first.
    ```bash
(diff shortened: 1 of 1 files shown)


_articles/appdev-deploy.md
```diff
@@ -142,7 +142,11 @@ Staging used to be deployed by this process, but this was changed to deploy the
    ```bash
    cd identity-devops
    ```
-1. Recycle the IDP instances to get the new code, it automatically creates a new migration instance first.
+1. Check current server status, and confirm that there aren't extra servers running. If there are, scale in old instances before deploying.
+   ```bash
+   aws-vault exec prod-power -- ./bin/ls-servers -e prod
```
zachmargolis, line 147: another option is `asg-size` which tell us the target/expected sizes, and if a resize is already scheduled?
```suggestion
   aws-vault exec prod-power -- ./bin/ls-servers -e prod
   aws-vault exec prod-power -- ./bin/asg-size prod idp
```

zachmargolis (approved):
LGTM

soniaconnolly:
> I seem to recall some existing or desired Slack alerting when the health pool size had been increased more time than what's reasonable.
> 
> Maybe a huddle discussion?

Adding this as an eng huddle topic sounds like a good idea.

_articles/appdev-deploy.md
soniaconnolly: Saw this after merging! That's also a good idea. I can make a new PR.

jmhooper:
This is a test comment for notification purposes.

PR #469, opened 2023-08-29
LG-10752 Cloudwatch tips for 500 errors, and show how to find deploy announce workflow in Slack

soniaconnolly:
Tips for finding out more about a 500 error in Cloudwatch, intended for appdev team oncall.

Also add an image showing how to find the deploy announcement workflow in Slack.


_articles/appdev-troubleshooting-production.md
```diff
@@ -126,3 +136,27 @@ We don't have indexed lookups by phone number so we need to combine a few approa
    | filter name = 'Telephoy: OTP sent'
    | filter properties.event_properties.phone_fingerprint in ["aaa", "bbb", "ccc"] # CHANGE THIS
    ```
+
+### Find the user id for a 500 error in production.log
+
+Click on the name of the error in NewRelic to get a more detailed view.
+At the top of that page, find the approximate time of the error. Note timezones for both NewRelic and AWS and convert if needed.
```
zachmargolis, line 143: we attach user_id as a custom property in NewRelic as well, but looking at the current crop of errors, I don't think it applies to background job errors, where most of our proofing errors come from

zachmargolis (approved):

_articles/appdev-troubleshooting-production.md
soniaconnolly: That sounds cool! I'm looking at [this error](https://one.newrelic.com/nr1-core/errors-inbox/entity-inbox/MTM3NjM3MHxBUE18QVBQTElDQVRJT058NTIxMzY4NTg?account=1376370&duration=2592000000&state=f52caf37-ba59-1cc4-773a-58c26126999e) and don't see it. Where should I be looking?

zachmargolis: In the Attributes section, like `amzn_trace_id` is a custom attribute, but 
Screenshot 2023-08-29 at 11 46 48 AM

but looking at the source code, I guess we no longer add in the user ID

https://github.com/18F/identity-idp/blob/619d748296b874d090d8ede0cd18bc3ad768ee70/app/controllers/application_controller.rb#L111-L115

or it gets added in `analytics.rb` but maybe there's a timing race or something

https://github.com/18F/identity-idp/blob/619d748296b874d090d8ede0cd18bc3ad768ee70/app/services/analytics.rb#L35-L36

soniaconnolly: No, you're right, when I scroll all the way down, I see it. I'll add a note about that.
Screenshot 2023-08-29 at 12 03 55 PM

PR #476, opened 2023-10-26
Adds text warning to check sha values before proceeding to recycle

Jeremy1026:
Adds the following text to Step 2's server check:

> Also ensure that there aren't multiple sha values, this would indicate that there is another recycle still in progress.


Jeremy1026:
Not actually needed. There is already a safety check in place.

PR #477, opened 2023-11-01
Update platform-gitlab.md

nick-soutouras:
Remove reference to Face/Touch Unlock

### DIFF
--- a/_articles/platform-gitlab.md
+++ b/_articles/platform-gitlab.md
@@ -47,7 +47,6 @@ Prerequisites:
 * You must use a phishing resistant multi-factor option for MFA which is one of:
   * PIV or CAC
   * Security Key (WebAuthN with a hardware key)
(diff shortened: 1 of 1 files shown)


aduth:
Can you share some context for this change? My understanding is that the text is accurate as-is.

timothy-spencer (approved):
yay!

nick-soutouras:
@aduth To my knowledge, Biometric is not currently an approved method for accessing Gitlab.  Unless something has changed in the last couple of weeks it is just PIV or Yubikey.

aduth:
Okay, I stumbled my way through some relevant Slack discussions and found this to be related to some compliance requirements. I think we could probably do a better job of describing the preceding sentence "You must use" to describe the actual requirements since Face/Touch is a valid phishing-resistant method, but not a blocker 🤷

PR #491, opened 2023-12-06
Linkify Google Calendar text

zachmargolis:
Make it easier to click!

### DIFF
--- a/assets/js/private-links.tsx
+++ b/assets/js/private-links.tsx
@@ -11,6 +11,7 @@ export const loadPrivateEye = () => {
       "calendar.gsa.gov",
       "cm-jira.usa.gov",
       "connect.gsa.gov",
+      "calendar.google.com",
       "docs.google.com",
       "drive.google.com",
       "ea.gsa.gov",
--- a/_articles/onboarding.md
+++ b/_articles/onboarding.md
@@ -31,7 +31,7 @@ Each Login.gov team has their own personalized [GDoc/Onboarding Template / Check
 - Request access to relevant [Google Groups](https://groups.google.com/).
 - Add the Login.gov Shared Calendar to your Google Calendar:
     - First make sure you are added to our core federal or contractor Google Group.
-    - Go to https://calendar.google.com/calendar/render?cid=gsa.gov_6ovul6pcsmgd40o8pqn7qmge5g@group.calendar.google.com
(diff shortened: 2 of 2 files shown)


assets/js/private-links.tsx
```diff
@@ -11,6 +11,7 @@ export const loadPrivateEye = () => {
       "calendar.gsa.gov",
       "cm-jira.usa.gov",
       "connect.gsa.gov",
+      "calendar.google.com",
```
aduth, line 14: Is this a public link? 🤔

aduth (approved):

assets/js/private-links.tsx
aduth: As I understand it, this configuration is for links which should be considered public.

zachmargolis: I think the config an name is weird, these are the URLs we do want the lock next to [docs link](https://github.com/18F/private-eye#:~:text=In%20the%20example%20above%2C%20all%20URLs%20matched%20by%20ignoreURLs%20will%20have%20the%20customized%20defaultMessage%20as%20the%20message%20the%20user%20sees%20when%20they%20hover%20over%20the%20link.)

But I am not sure why the lock isn't appearing on this page

aduth: I see two different initializations of PrivateEye. Maybe we updated the wrong one? (Why are there two?)

https://github.com/18F/identity-handbook/blob/eaf5150315a80b963fa13ba4c57daf043ff88232/assets/js/setup.ts#L14

https://github.com/18F/identity-handbook/blob/eaf5150315a80b963fa13ba4c57daf043ff88232/assets/js/private-links.tsx#L3

aduth: Alternatively / additionally, I see an existing entry for calendar.gsa.gov above, so we could update the link in the content (I don't see an actual link elsewhere to calendar.gsa.gov , but it does seem to redirect correctly).

zachmargolis: Updated in 54e031c6f557fcf9949512861a96dc35864ef166 by removing extra copy, must have been an incomplete refactor.

Confirmed the lock icon appears, I think I may give the link a title so it doesn't overflow
Screenshot 2023-12-06 at 11 30 48 AM

aduth (approved):
LGTM! Nice bit of cleanup too

PR #492, opened 2023-12-14
LG-11797 FlowPolicy documentation

soniaconnolly:
Documentation for new FlowPolicy, which manages which steps are allowed for an Identity Verification session and user.

Appdev - Architecture - [FlowPolicy page](https://federalist-7d0b2b76-42fc-4df1-9825-8c3cd77e0a15.sites.pages.cloud.gov/preview/18f/identity-handbook/sonia-flow-policy-documentation/articles/identity-verification-flow-policy.html)

### DIFF
--- a/_articles/identity-verification-flow-policy.md
+++ b/_articles/identity-verification-flow-policy.md
@@ -21,10 +21,19 @@ This is how individual Identity Verification controllers implement the FlowPolic
 
 [FlowPolicy code](https://github.com/18F/identity-idp/blob/main/app/policies/idv/flow_policy.rb)
 
-The FlowPolicy contains a private hash of steps. The new controller's key and step_info need to
-be added to the hash.
+FlowPolicy contains a private hash of steps. The new controller's key and step_info need to
+be added to the hash. For example: `agreement: Idv::AgreementController.step_info,`
 
-For example: `agreement: Idv::AgreementController.step_info,`
+FlowPolicy has three public methods. They are called from helper methods in IdvStepConcern ([see below](#idvstepconcern-methods)) rather than being used directly.
+
+`controller_allowed?` uses the controller's `StepInfo.preconditions` to determine if the user is allowed to visit the given controller.
+
+`info_for_latest_step` returns the `StepInfo` object for the latest step the user is allowed to be on. The `StepInfo` object contains enough information to build a url to redirect the user.
+
+`undo_future_steps_from_controller!` uses the controller's `StepInfo.next_steps` to find all future steps and tail-recursively call `undo_step` on them.
+
+The FlowPolicy contains a private hash of steps. The new controller's key and step_info need to
+be added to the hash. For example: `agreement: Idv::AgreementController.step_info,`
 
 ## StepInfo
 
@@ -66,12 +75,12 @@ This general before action replaces specific before actions that check that the 
 deleted. Before actions that check if this step is needed would prevent the back button from 
 working, so they should be deleted.
 
+### `clear_future_steps!`
+
 The first line in `#update` (or `#create` for a few controllers) should call `clear_future_steps!`.
 When the user submits a step, their session is reset to invalidate all downstream steps. This is
 important for security to make sure someone cannot verify one set of data and then change parts
 of it afterward.
-
-### `clear_future_steps!`
 
 `clear_future_steps!` looks at `step_info.next_steps` for the current step and calls
 `step_info.undo_step` for all future steps, using tail recursion to traverse the tree of steps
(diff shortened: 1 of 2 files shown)


_articles/identity-verification-flow-policy.md
```diff
@@ -0,0 +1,109 @@
...
+`FlowPolicy::FINAL`, for example PhoneErrorsController.step_info. It can be tricky to determine
+where a step belongs. For example :request_letter is in the next_steps array for VerifyInfo
+because even though it is accessed from the Phone step, the Phone step does not have to be
+submitted to reach RequestLetter.
+* preconditions - a proc with arguments `(idv_session:, user:)` that returns true if a
+user is allowed to be on this step
+* undo_step - a proc with arguments `(idv_session:, user:)` that clears any fields set by
+this step. Called from `clear_future_steps!`
+* action - a symbol for the controller action to use if jumping to this step. Defaults to `:show`,
+so it is only needed for controllers that use `:new` or `:index`.
```
solipet, line 44: can we highlight the fields?
```suggestion
* `key` - a unique key which is a symbol, for example `:agreement`
* `controller` - self, which is converted to the full controller "path" by `initialize`, for example
`/idv/agreement`
* `next_steps` - an array of keys of following steps once this step is submitted. Some controllers
only have one. Agreement has
several: `[:hybrid_handoff, :document_capture, :how_to_verify]`. If there is no following step, use
`FlowPolicy::FINAL`, for example PhoneErrorsController.step_info. It can be tricky to determine
where a step belongs. For example :request_letter is in the next_steps array for VerifyInfo
because even though it is accessed from the Phone step, the Phone step does not have to be
submitted to reach RequestLetter.
* `preconditions` - a proc with arguments `(idv_session:, user:)` that returns true if a
user is allowed to be on this step
* `undo_step` - a proc with arguments `(idv_session:, user:)` that clears any fields set by
this step. Called from `clear_future_steps!`
* `action` - a symbol for the controller action to use if jumping to this step. Defaults to `:show`,
so it is only needed for controllers that use `:new` or `:index`.
```

_articles/identity-verification-flow-policy.md
```diff
@@ -0,0 +1,109 @@
...
+deleted. Before actions that check if this step is needed would prevent the back button from 
+working, so they should be deleted.
+
+The first line in `#update` (or `#create` for a few controllers) should call `clear_future_steps!`.
+When the user submits a step, their session is reset to invalidate all downstream steps. This is
+important for security to make sure someone cannot verify one set of data and then change parts
+of it afterward.
+
+`clear_future_steps!` looks at `step_info.next_steps` for the current step and calls
+`step_info.undo_step` for all future steps. It does not call `undo_step` for the current step,
```
solipet, line 60: Worth pointing out that `clear_future_steps!` works in reverse from the end.

```suggestion
`clear_future_steps!` looks at `step_info.next_steps` for the current step and calls
`step_info.undo_step` for all future steps, working from the last step back to the directly
adjacent next steps. It does not call `undo_step` for the current step,
```

_articles/identity-verification-flow-policy.md
```diff
@@ -0,0 +1,109 @@
...
+      subject.idv_session.applicant = Idp::Constants::MOCK_IDV_APPLICANT
+      expect(subject).to receive(:clear_future_steps!).and_call_original
+
+      put :update
+      expect(subject.idv_session.applicant).to be_nil
+    end
+```
+
+Because `url_for_latest_step` stops at the first step whose conditions are not met, controller
+specs now have to have `idv_session` fully set up for the step under test. Use the `stub_step`
```
solipet, line 102: Since the example uses `stub_up_to` lets just mention that.

```suggestion
specs now have to have `idv_session` fully set up for the step under test. Use the `stub_up_to`
```

_articles/identity-verification-flow-policy.md
```diff
@@ -0,0 +1,109 @@
...
+helper in FlowPolicyHelper with the key of the previous step. For example, 
+`stub_up_to(:ssn, idv_session: idv_session)` prepares the session for testing in
+verify_info_controller_spec.
+
+# idv_session 
+mark step complete
+invalidate step
```
solipet, line 109: Should we flesh this out?

solipet (changes requested):
A few suggestions (not required). The section on `idv_session` didn't get fleshed out - lets do that before merging.

solipet (approved):
Yay!

theabrad (approved):
Very nice write-up! LGTM

matthinz (approved):
Excellent work!

PR #493, opened 2023-12-15
New article on how to manage the 50/50 state

soniaconnolly:
During deploys, both new and old instances are serving requests. This is called the 50/50 state and requires careful management when changing code that is used across instances.

Appdev - Development - [How to Manage the 50/50 State](https://federalist-7d0b2b76-42fc-4df1-9825-8c3cd77e0a15.sites.pages.cloud.gov/preview/18f/identity-handbook/sonia-50-50-state/articles/manage-50-50-state.html)

### DIFF
--- a/_articles/manage-50-50-state.md
+++ b/_articles/manage-50-50-state.md
@@ -15,12 +15,13 @@ are serving requests for at least 15 minutes. This is called the 50/50 state and
 management when changing code that is used across instances.
 
 You can test the 50/50
-state locally by loading a page on your new branch, killing the server, checking out main, and then
-clicking on a link on the branch (new instance), which will load the next action on main (old
+state locally by loading a page on your new branch, killing the server, checking out main,
+restarting the server, and then
+clicking on a link on the page (new instance), which will load the next action on main (old
 instance). For more complex
 changes, you may want to [test in the 50/50 state](#test-in-the-5050-state) in a lower environment.
 
-The cases below sound similar, but each situation has its own considerations.
+The cases described below sound similar, but each situation has its own considerations.
 
 ## Database changes
 
--- a/_articles/manage-50-50-state.md
+++ b/_articles/manage-50-50-state.md
@@ -14,8 +14,13 @@ During deploys, we add new instances before shutting down old ones, so both new 
 are serving requests for at least 15 minutes. This is called the 50/50 state and requires careful
 management when changing code that is used across instances.
 
-The cases below sound similar, but each situation has its own considerations. For more complex
-changes, you may want to [test in the 50/50 state](#test-in-the-5050-state).
+You can test the 50/50
+state locally by loading a page on your new branch, killing the server, checking out main, and then
+clicking on a link on the branch (new instance), which will load the next action on main (old
+instance). For more complex
+changes, you may want to [test in the 50/50 state](#test-in-the-5050-state) in a lower environment.
+
+The cases below sound similar, but each situation has its own considerations.
 
 ## Database changes
 
@@ -23,31 +28,37 @@ changes, you may want to [test in the 50/50 state](#test-in-the-5050-state).
 * Deploy 1: Add the migration, write to the new table/field. Optionally add a rake task to update
 existing data.
 * After deploy (optional): Run the rake task to update existing data.
-* Deploy 2: Read from the table/field.
+* Deploy 2: Read from the table/field
 
-## Persistent data structures
+### Remove a table or field
+* Deploy 1: Remove all reads in the code
+* Deploy 2: Remove all writes in the code
+* Deploy 3: Remove the table/field
+
+## Persistent data structures in redis
 
 When adding, removing, or renaming fields from a data structure that is persisted in redis, such as
 `user_session`, `idv_session`, or `document_capture_session`, it could be saved from a new instance
 and read by an old one, or vice versa.
 
 ### Add a field
-* Deploy 1: Write to the new field but do not use it
+* Deploy 1: Add the field and write to it
 * Deploy 2: Read from the field
 
 ### Remove a field
 * Deploy 1: Remove everywhere that reads the field
-* Deploy 2: Remove the field
+* Deploy 2: Remove all writes to the field
+* Deploy 3: Remove the field
 
 ### Rename a field
-* Deploy 1: Write to the new name everywhere the old name is written. Read from both fields with
+* Deploy 1: Add the new name. Write to the new name everywhere the old name is written. Read from both fields with
 `new_name || old_name`
 * Deploy 2: Remove references to the old name
 
 ## Jobs
 
 Job requests could be enqueued by a new instance and picked up by an old one, or vice versa. When an
-argument is added or removed, it causes 500 errors from ArgumentError.
+argument is added or removed, it causes 500 errors with ArgumentError.
 
 ### Add an argument to a job #perform method
 * Deploy 1: Add the argument with a default value (usually nil) and do not use it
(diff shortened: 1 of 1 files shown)


_articles/manage-50-50-state.md
```diff
@@ -0,0 +1,138 @@
...
+* Deploy 2: Remove references to the old name
+
+## Jobs
+
+Job requests could be enqueued by a new instance and picked up by an old one, or vice versa. When an
+argument is added or removed, it causes 500 errors from ArgumentError.
+
+### Add an argument to a job #perform method
+* Deploy 1: Add the argument with a default value (usually nil) and do not use it
+* Deploy 2: Use the argument, and gracefully handle the default value
```
aduth, line 54: Do these need to be separate deploys? 🤔

_articles/manage-50-50-state.md
```diff
@@ -0,0 +1,138 @@
...
+When deleting a route, an old instance could request the route, and a new instance would not
+recognize it.
+
+### Add a route
+* Deploy 1: Add the fully functional route and controller action
+* Deploy 2: Link and redirect to the route as needed
+
+### Delete a route
+* Deploy 1: Remove all links and redirects to the route
+* Deploy 2: Change the route to a redirect if possible to avoid user 404s from bookmarks.
```
aduth, line 77: Most of our routes are authenticated and we don't "remember" destinations (LG-3737) so I wouldn't expect bookmarks to be very useful in the first place.

_articles/manage-50-50-state.md
```diff
@@ -0,0 +1,138 @@
...
+* Deploy 3: 2-4 weeks later, check logs and remove the deprecated route.
+
+### Rename a route
+* Deploy 1: Add the new route and controller action. Make the old route point to the new controller.
+* Deploy 2: Delete the old controller action.
+* Deploy 3: 2-4 weeks later, check logs and remove the deprecated route.
```
aduth, line 83: Maybe we can mention the specific logs to check?
```suggestion
* Deploy 3: 2-4 weeks later, check `production.log` request logs and remove the deprecated route.

### Rename a route
* Deploy 1: Add the new route and controller action. Make the old route point to the new controller.
* Deploy 2: Delete the old controller action.
* Deploy 3: 2-4 weeks later, check `production.log` request logs and remove the deprecated route.
```

_articles/manage-50-50-state.md
```diff
@@ -0,0 +1,138 @@
...
+* Deploy 1: Add the fully functional route and controller action
+* Deploy 2: Link and redirect to the route as needed
+
+### Delete a route
+* Deploy 1: Remove all links and redirects to the route
+* Deploy 2: Change the route to a redirect if possible to avoid user 404s from bookmarks.
+* Deploy 3: 2-4 weeks later, check logs and remove the deprecated route.
+
+### Rename a route
+* Deploy 1: Add the new route and controller action. Make the old route point to the new controller.
```
aduth, line 81: For the second sentence, would this be a redirect, or a normal route mapping?

_articles/manage-50-50-state.md
```diff
@@ -0,0 +1,138 @@
...
+
+* Deploy 1: Change the flag value
+* Make any of the above changes that are needed in following deploys
+
+## Test in the 50/50 state
+When making a large change such as adding a new page to the Identity Verification flow, it can be
+helpful to test in the 50/50 state. Large changes are developed behind a feature flag and then
+turned on.
+
+To create a long-running 50/50 state in int:
```
aduth, line 100: Should we use `dev` for this kind of testing to avoid creating issues for partners in the integration environment?

_articles/manage-50-50-state.md
```diff
@@ -0,0 +1,138 @@
...
+You will see old instances that will start to say draining
+You will know it is ready when you only have new workers and idp instances up- check the time- should be minutes
+1. Set your feature flag to true
+```
+aws-vault exec sandbox-power -- ./bin/app-s3-secret --env int --app idp --edit
+```
+Add the flag in alphabetical order 
+ESC, :wq to exit vim
+Examine diff, y to save changes
+copy/paste diff to Slack thread about recycle in #login-appdev
```
aduth, line 120: Not sure if it's expected, but the newline formatting doesn't carry forward into the post-Markdown render.

_articles/manage-50-50-state.md
```diff
@@ -0,0 +1,138 @@
...
+1. Double the number of instances
+This gives you half new instances with the flag set to true, and
+half old instances with the flag set to false.
+```
+aws-vault exec sandbox-power -- ./bin/asg-size int idp # (outputs current number of idp instances x)
+aws-vault exec sandbox-power -- ./bin/asg-size int idp 2x
```
aduth, line 126: I'd worry someone might actually type "2x" in this command rather than substituting the output from the previous value. Maybe we could be more clear that it's meant to be substituted, or just hard-code a value based on what we know to be the typical size in sandbox environments?

aduth (approved):
Handful of minor notes, but LGTM overall 👍  Will be great to have an article to point to for something that's frequently brought up in conversation.

_articles/manage-50-50-state.md
soniaconnolly: I just [messed this up yesterday](https://gsa-tts.slack.com/archives/C0NGESUN5/p1702583900546089), so I think they do have to be separate deploys. Maybe @jmhooper can explain it better.

_articles/manage-50-50-state.md
soniaconnolly: Wow that ticket has been around for a while! People could use bookmarks/history after they log in. We occasionally see 500s from jumps like that when we forget to take that into account.

_articles/manage-50-50-state.md
matthinz: Dev redeploys on every commit to main, so it's trickier to keep a consistent 50/50 state active there. This might be a case for team (or personal) sandboxes if available?

_articles/manage-50-50-state.md
matthinz: I noted this as well. It would be cool to have asg-size acccept "2x" as a valid arg 😄

matthinz (approved):
This is so great. There are a few minor markdown line breaking issues that you can see in the preview, but this info is so incredibly valuable. Thank you for doing this!

_articles/manage-50-50-state.md
zachmargolis: yes maybe we should add an explainer sentence why. Jobs are harder because we have a lil 2x2 matrix here:

```
{ enqueued by old, new } x { performed on old, new }
```
The big cases to watch for
- new arguments need to be optional because old boxes won't enqueue them
- removed arguments need to stick around for one deploy as optional because old boxes will still enqueue them anyways

zachmargolis: nevermind I see that we do have an explainer para up above saying this

soniaconnolly: I clarified between not using the argument inside perform (which is what I said) and not using it when calling perform (which is what I need to say). Thanks for pointing out this was unclear/unfinished.

_articles/manage-50-50-state.md
soniaconnolly: Added more commentary about what environment to use.

_articles/manage-50-50-state.md
soniaconnolly: @aduth thanks for your comments! Addressed them in additional commits.

_articles/manage-50-50-state.md
soniaconnolly: Updated to use actual numbers from int.

PR #504, opened 2024-01-23
update README with additional instructions for running locally

jgrevich:
Title says it all. This is part of the follow up from the [incident review](https://docs.google.com/document/d/1j5ihCdAZUssQN9Y9KevvOEqFcBXU48Iz_66JF9yVE28/edit#heading=h.yjzagco8hnao) for the login.gov brochure outage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,17 @@ article and it will be automatically linked to from the index of all articles.
 
 ## Running Locally
 
+Ensure that you have the concurrently CLI helper installed.
+
+```
+npm i -g concurrently
+```
+
 ```
 make run
 ```
+
+Open http://127.0.0.1:4000/ in a browser.
 
 ## Running Tests
 
--- a/README.md
+++ b/README.md
@@ -11,17 +11,17 @@ article and it will be automatically linked to from the index of all articles.
 
 ## Running Locally
 
-Ensure that you have the concurrently CLI helper installed.
+Ensure that you have the required dependencies installed.
 
 ```
-npm i -g concurrently
+make setup
 ```
 
 ```
 make run
 ```
 
-Open http://127.0.0.1:4000/ in a browser.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,10 +11,18 @@ article and it will be automatically linked to from the index of all articles.
 
 ## Running Locally
 
+Ensure that you have the concurrently CLI helper installed.
+
+```
+npm i -g concurrently
+```
+
 ```
```
zachmargolis, line 20: I think we just need to make sure all node packages are installed? concurrently is in our package.json
~~~suggestion

```
make setup
~~~

File: README.md [hunk 2]
```diff
@@ -11,10 +11,18 @@ article and it will be automatically linked to from the index of all articles.
...
 make run
 ```
 
+Open http://127.0.0.1:4000/ in a browser.
```
jmdembe, line 24: `http//localhost:4000` can get you to the app as well.

jgrevich: Thanks. I’ll update with localhost which is likely preferred over 127.0.0.1.

README.md
jgrevich: I tried that at first but that doesn’t seem to install the concurrently CLI tool. The only way I could get that working was with the above command. I’m not super familiar with node packaging so maybe I’m missing something.

aduth: I think @zachmargolis's suggestion of `make setup` would be a necessary addition and would include `npm install`, I could see how the current instructions may not always work if you didn't have all the packages installed.

Maybe worth checking if it works for you with that flow?

```
npm uninstall -g concurrently
make setup
make run
```

jgrevich: Odd, I can't reproduce the problem. Uninstalling and running `make setup` works fine. I'll update the readme to run `make setup` instead of installing `concurrently` globally. Thanks for pointing that out.

jgrevich:
This has been rebased from main and I all comments and suggestions have been addressed.

zachmargolis (approved):

PR #508, opened 2024-02-05
Add wargames player page

Sgtpluck:
This change adds a wargames player's guide. This includes:
- A tech lead preparation list

(Ideally, we could build out other prep lists for other roles, and other helpful information that may relieve anxiety in volunteering for a new role.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "identity-handbook",
       "dependencies": {
         "@18f/identity-build-sass": "^1.1.0",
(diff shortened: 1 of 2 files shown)


package-lock.json
aduth, line 1: Are these changes expected?

mitchellhenke (approved):

package-lock.json
Sgtpluck: when i set up my repo, i got an npm audit warning, so i fixed it. happy to revert that commit if that wasn't the right thing!

aduth: I think they'd be alright to keep if they were intentional 👍

PR #512, opened 2024-02-10
LG-12244: Add link to LexisNexis TrueID outage runbook

night-jellyfish:
[LG-12244](https://cm-jira.usa.gov/browse/LG-12244)

Added a link to the new LexisNexis TrueID outage runbook.

Also added the other possibility for `VENDOR_STATUS_OPTIONS`, [which I found](https://github.com/18F/identity-idp/blob/cf8764c2be19a834c3c7116d35bfccb44de41938/lib/identity_config.rb#L6) in my research for creating the runbook.

I didn't find any behavioral difference between `full_outage` and `partial_outage` in my local testing, though I could've missed something.

### DIFF
--- a/_articles/vendor-outage-response-process.md
+++ b/_articles/vendor-outage-response-process.md
@@ -77,17 +77,16 @@ individually. Each is controlled by a configuration flag:
 The possible values for each flag:
 
 - `operational`
-- `partial_outage`
 - `full_outage`
 
 The default value for each of the flags is `operational`.
 
-When one or more of the flags are set to `full_outage` or `partial_outage`, some parts of
+When one or more of the flags are set to `full_outage`, some parts of
 identity verification will be disabled.
 
 As an overview:
 
-- Setting `full_outage` or `partial_outage` for `acuant`, `lexisnexis_instant_verify`, or
+- Setting `full_outage` for `acuant`, `lexisnexis_instant_verify`, or
   `lexisnexis_trueid` turns off pretty much everything. Identity verification
   is completely unavailable.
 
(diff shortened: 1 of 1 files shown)


night-jellyfish:
@aduth when I tested locally, I saw the same behavior as `full_outage` with `partial_outage`. I think because [our check for an outage](https://github.com/18F/identity-idp/blob/889e7a6e351ac93e35493bf6fb0b255f88336176/app/services/outage_status.rb#L32) just looks for `!= operational`.

There could be more nuanced behavior that I didn't notice though. With that in mind do you still want to remove it? I suppose if I don't fully understand what it all does, maybe it's not good to document it. It just caused me a little confusion to see it as an option in the constant with no documentation about it.

aduth:
I think we should leave it out, until it's implemented to have a meaningful difference. Otherwise I'd worry its presence could cause confusion in the midst of an incident.

It looks like it was documented previously and intentionally removed in #401 by @matthinz  for this reason:

>(I since I was in there, I also removed references to `partial_outage`--support for this was not implemented.)

night-jellyfish:
I see! Thanks for that context @aduth! I was thinking of not implemented as "not doing anything" but really it's, "doesn't do what it was intended to do." I wonder if it's worth a ticket to remove it entirely? I am not sure who would own that work or if there is other work behind the scenes to implement `partial_outage` in the future. Sounds like perhaps @matthinz might know?

For now I'm going to:

- push up a change to remove the references to it I made here
- rebase
- merge if CI passes

aduth:
>I wonder if it's worth a ticket to remove it entirely?

That's a good question. Personally, I think it's worth removing. If I recall, it was initially implemented as part of a greater effort to have granular support for vendor outages, but that effort was only half-completed before it was deprioritized in favor of other work. And that was several years ago, so if we haven't revisited it by now, we may as well remove whatever defunct code we half-implemented.

PR #515, opened 2024-02-21
Remove design.login.gov reference from DOD

nickttng:
**What:**
I came across a `design.login.gov` reference while reviewing the definition of done page. 

Removing since it doesn't exist anymore while replacing it with a link to the the LGDS repository.


aduth (approved):
Good catch! This prompted me to search for any other instances in the repository and I found another, maybe we should update that as well?

https://github.com/18F/identity-handbook/blob/05fab433d71f9a5dfc00a359a1d182a78e8b5dab/_articles/onboarding.md#L24

nickttng:
@aduth Oh, yes - I updated that page via https://github.com/18F/identity-handbook/pull/515/commits/cc0426c73c21fca0158f14ed669a4440f525feab

aduth (approved):
LGTM!

PR #516, opened 2024-02-22
Update piv-ssh.md

mzia:
resolved to point to updated URL


aduth (approved):
Hey @mzia 👋 😄  LGTM! 

Aside: Might be a good idea to have an external links checker similar to [what we have for the main login.gov brochure site](https://github.com/18F/identity-site/blob/62555ac3373bfbb8730dc4cd79c4d1452c8f9a5b/.circleci/config.yml#L160-L176).

PR #521, opened 2024-03-01
Fix typo for identity-logging repository

aduth:
Previous spelling "identity-loging" was both mispelled and resulted in a 404 for the link navigation.


aduth:
While I'm here, maybe I should flag this repository as archived, since I don't think we use it?

zachmargolis (approved):
LGTM and yes! we do use it currently

PR #531, opened 2024-03-26
LG-12547 Updated GPO letter image.

jmax-gsa:
We've changed the text on the GPO letter slightly; this is an update to the image of a GPO letter so it matches what we're currently sending.


aduth (approved):
👍 

The file size is pretty big (3.5MB). I might suggest either resizing it to be slightly smaller and/or running it through an image optimizer like [Squoosh](https://squoosh.app/).

aduth (approved):
LGTM 👍 Thanks for optimizing the image

aduth:
Projects in the GSA-TTS organization require commit signing, so you'll need to set that up and rebase/amend your commits to be "Verified".

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #541, opened 2024-05-03
Private Handbook Flip-Flop

zachmargolis:
- Removes all private login parts
- Adds link to private handbook on repos page (and mentioned Gitlab)

Related PR: https://gitlab.login.gov/lg-public/identity-internal-handbook/-/merge_requests/7


_articles/github.md
```diff
@@ -129,9 +131,12 @@ View a [list of all repositories](https://github.com/topics/login-gov) tagged fo
   Hosted on: [Cloud.gov Pages](https://pages.cloud.gov/)<br />
   This handbook!
 
-- [**`18f/identity-handbook-private`**](https://github.com/18f/identity-handbook-private) <br />
-  [login-handbook.app.cloud.gov](https://login-handbook.app.cloud.gov/)<br />
-  Hosted on: [Cloud.gov](https://dashboard.fr.cloud.gov/)<br />
+- [**`lg-public/identity-internal-handbook`**](https://gitlab.login.gov/lg-public/identity-internal-handbook/)<br />
```
aduth, line 134: Maybe a future enhancement, but it'd be nice to be able to easily distinguish which of these links are GitHub vs. GitLab. I guess the 🔒 is a pretty obvious indicator, though I think I'd also expect that to be shown for private GitHub repositories 🤷  Something like an icon could be a cool future enhancement.

assets/js/private-articles.tsx
```diff
@@ -1,237 +0,0 @@
-import { render, Fragment, createContext, ComponentChildren } from "preact";
-import { useEffect, useMemo, useContext } from "preact/hooks";
-import { load as loadYAML } from "js-yaml";
```
aduth, line 3: This is an unused dependency now we could remove from `package.json`.

package.json
```diff
@@ -6,12 +6,6 @@
     "anchor-js": "^4.2.2",
     "concurrently": "^8.0.1",
     "esbuild": "^0.17.15",
-    "htm": "^3.1.0",
-    "js-yaml": "^4.1.0",
-    "marked": "^4.0.12",
```
aduth, line 11: We should also remove `@types/marked`.

tsconfig.json
```diff
@@ -3,13 +3,11 @@
     "strict": true,
     "esModuleInterop": true,
     "noEmit": true,
-    "jsx": "react-jsx",
```
aduth, line 6: If we're dropping all JSX support, we could also:

1. Remove [React ESLint overrides](https://github.com/GSA-TTS/identity-handbook/blob/5e102ee51a80b6c4257384d6706d96bbe41460e7/.eslintrc#L20-L22)
2. Remove [optional JSX peer dependency for shared ESLint configuration](https://github.com/GSA-TTS/identity-handbook/blob/5e102ee51a80b6c4257384d6706d96bbe41460e7/package.json#L26)
3. Remove [`.jsx` and `.tsx` extensions from linting](https://github.com/GSA-TTS/identity-handbook/blob/5e102ee51a80b6c4257384d6706d96bbe41460e7/package.json#L48)

_articles/github.md
zachmargolis: Good idea! Done in 46fe6c34f772124a553b379ff7c92b87be4c89cf

Demo:
Screenshot 2024-05-03 at 10 56 57 AM

assets/js/private-articles.tsx
zachmargolis: Yup, already done in ea410498d166561def43cdbd03c858ca36f92f14

tsconfig.json
zachmargolis: done in 7bc04f7308fb043d2004527db1ec34dd0852146a!

package.json
zachmargolis: Done in 7bc04f7308fb043d2004527db1ec34dd0852146a

assets/js/private-articles.tsx
aduth: Ah I think my searching was revealing some transitive dependencies in `package-lock.json`, or maybe the DefinitelyTyped package you removed in 7bc04f7308fb043d2004527db1ec34dd0852146a.

aduth (approved):
LGTM 👍

PR #546, opened 2024-05-14
Remove text stating updates can be made by whatever means necessary

Jeremy1026:
Currently the page regarding [Secrets and Configuration](https://handbook.login.gov/articles/appdev-secrets-configuration.html) includes language implying that updates can be made by editing the secrets file in s3 by downloading, editing, and uploading. This is not recommended, users should use the `app-s3-secret` script to make edits. So this PR removes that language.

[Conversation in Slack regarding this change](https://gsa.enterprise.slack.com/archives/C16RSBG49/p1715694023104519)

### DIFF
--- a/_articles/appdev-secrets-configuration.md
+++ b/_articles/appdev-secrets-configuration.md
@@ -21,10 +21,6 @@ environment by merging default values with an environment-specific YAML file.
 The S3 buckets that contain secrets are versioned, so we can recover old versions
 if needed.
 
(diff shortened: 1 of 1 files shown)


Sgtpluck (approved):
thanks!

ajfarkas (commented):
Would it make sense to include instructions to use `app-s3-secret`?

Jeremy1026:
> Would it make sense to include instructions to use `app-s3-secret`?

There is instruction for how to use the script directly below.

PR #550, opened 2024-05-21
Update Team Agnes focus

zachmargolis:


jc-gsa (approved):
Looks good. The syntax/style for the value of the key focus_area seems to differ a little across teams

zachmargolis:
Yes! Probably worth fixing eventually

PR #553, opened 2024-05-21
LG-13016: v2 remove oncall by request

jc-gsa:
[LG-13016](https://cm-jira.usa.gov/browse/LG-13016)

This change removes the oncall requirement for Team Annie by request.


zachmargolis (approved):
LGTM, you could also remove `product: true` so that the requirement doesn't apply? It doesn't affect anything visually on the page

PR #556, opened 2024-05-22
Update offboarding.md

benjaminchait:
I believe we're moving out of GitHub for this, and into GitLab...?

cc @nick-soutouras


nick-soutouras (approved):
LGTM

PR #557, opened 2024-05-24
Update SAML Fingerprint process

ajfarkas:
The process for getting the certificate was a little too vague to follow.
I updated this to indicate that there are different processes for the hosted and local apps.

I also can't tell you why the local process is that way, but that's the only way I've been able to get a valid fingerprint. I'd love to have someone explain this.

### DIFF
--- a/_articles/saml.md
+++ b/_articles/saml.md
@@ -51,19 +51,9 @@ If, when logging in to the SAML Sinatra sample app, you get an error saying:
 
 This is usually caused by a mismatch between the IdP certificate used to sign the response, and the recorded signature of the certificate which is saved in the environment variable `idp_cert_fingerprint` (either in config/application.yml, or the environment variables in the deployed environment).
 
-To fix this, grab the certificate from the response, e.g.,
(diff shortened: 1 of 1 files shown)


_articles/saml.md
```diff
@@ -52,6 +52,7 @@ If, when logging in to the SAML Sinatra sample app, you get an error saying:
 This is usually caused by a mismatch between the IdP certificate used to sign the response, and the recorded signature of the certificate which is saved in the environment variable `idp_cert_fingerprint` (either in config/application.yml, or the environment variables in the deployed environment).
 
 To fix this, you'll first need to get the X509 Certificate from the appropriate SAML metadata endpoint, ie `https://idp.dev.identitysandbox.gov/api/saml/metadata2024` for `https://dev-identity-saml-sinatra.app.cloud.gov/`.
+The local `identity-saml-sinatra` app uses the certificate from `identity-idp`, so that XML can be found at `identity-idp/config/artifacts.example/local/saml2024.crt`.
```
zachmargolis, line 55: what if we linked these?
```suggestion
The local [`identity-saml-sinatra`](https://github.com/18F/identity-saml-sinatra/) app uses the certificate from [`identity-idp`](https://github.com/18F/identity-idp/), so that XML can be found [in its examples directory](https://github.com/18F/identity-idp/tree/main/config/artifacts.example/local).
```

zachmargolis (approved):

PR #560, opened 2024-05-30
Override GSA-TTS .allstar config for "dismiss stale reviews"

zachmargolis:
**Why**: We believe this still meets the AC-2 control, while also allowing for a more flexible workflow for Login.gov contributors

Based on a [Slack conversation in #tts-devtools](https://gsa-tts.slack.com/archives/C02U4Q9M5UH/p1717091155987799), it sounds like we might be able to override this

### DIFF
--- a/.allstar/branch_protection.yaml
+++ b/.allstar/branch_protection.yaml
@@ -1,0 +1,1 @@
+dismissStale: false


aduth (approved):
It's not clear to me if this file overrides everything contained in [the default configuration](https://github.com/GSA-TTS/.allstar/blob/main/branch_protection.yaml), or is merged with it? In other words, do we need to duplicate all the other configuration values in that file?

zachmargolis:
> It's not clear to me if this file overrides everything contained in [the default configuration](https://github.com/GSA-TTS/.allstar/blob/main/branch_protection.yaml), or is merged with it? In other words, do we need to duplicate all the other configuration values in that file?

I don't know! I'm inclined to merge it, remove the setting, and see if we get flagged for it?

aduth:
> > It's not clear to me if this file overrides everything contained in [the default configuration](https://github.com/GSA-TTS/.allstar/blob/main/branch_protection.yaml), or is merged with it? In other words, do we need to duplicate all the other configuration values in that file?
> 
> I don't know! I'm inclined to merge it, remove the setting, and see if we get flagged for it?

Sure 👍 I'd guess we'd need to remove some other setting we're still expecting to be enforced (like `requireApproval`) to really validate how it works?

zachmargolis:
I have unchecked the box, will see if we get nagged via GitHub issue

Screenshot 2024-05-30 at 12 28 38 PM

mitchellhenke:
> Sure 👍 I'd guess we'd need to remove some other setting we're still expecting to be enforced (like `requireApproval`) to really validate how it works?

Yeah, can we confirm we haven't disabled all the rules?

zachmargolis:
I tried disabling this, which should also be a policy violation, so we should get a nag about this

Screenshot 2024-05-30 at 12 48 23 PM

zachmargolis:
Ok, no nag, so oops, this disabled everything. I'll redo this to include the full configs from the parent repo.

aduth:
> Ok, no nag, so oops, this disabled everything. I'll redo this to include the full configs from the parent repo.

Do we know if we'd expect to receive a notification immediately, or is it checked on a recurring delay?

zachmargolis:
https://github.com/GSA-TTS/identity-handbook/issues/519 Got an error, I will re-enable the "do not allow admin bypass"

PR #562, opened 2024-05-30
Redo: Override GSA-TTS .allstar config

zachmargolis:
Copies in entire config so that we can override one line of it, follow-up to https://github.com/GSA-TTS/identity-handbook/pull/560

### DIFF
--- a/.allstar/branch_protection.yaml
+++ b/.allstar/branch_protection.yaml
@@ -1,7 +1,7 @@
 # Branch protection rules
-# Implements AC-2 (part g):  requires peer review by at least 
+# Implements AC-2 (part g):  requires peer review by at least
 #   one other team member for the production “main/master” branch.
-# AC-2 (part k): cans, checks, and branch protection policies are enforced 
+# AC-2 (part k): cans, checks, and branch protection policies are enforced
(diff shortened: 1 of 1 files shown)


.allstar/branch_protection.yaml
```diff
@@ -1 +1,20 @@
-dismissStale: false
+# Branch protection rules
+# Implements AC-2 (part g):  requires peer review by at least
+#   one other team member for the production “main/master” branch.
+# AC-2 (part k): cans, checks, and branch protection policies are enforced
+#   configurations through the GSA-TTS Github Allstar implementation.
+# SI-7 Software, Firmware, and Information Integrity
+optConfig:
```
mitchellhenke, line 7: Should we be concerned about drift if this means changes in the upstream rules won't be captured here?

aduth (approved):
Not entirely certain we need this pending https://github.com/GSA-TTS/identity-handbook/pull/560#issuecomment-2140773380, but would expect to be safe regardless

zachmargolis:
Closing, looks like we got a violation as expected: https://github.com/GSA-TTS/identity-handbook/issues/519#issuecomment-2140789666

that shows only one violation (instead of 2), so this is not needed

PR #575, opened 2024-06-21
Rename postmortem template to incident review template

zachmargolis:
**Why**: We have moved away from the term postmortem


nick-soutouras (approved):
lgtm

aanand-gsa (approved):

PR #584, opened 2024-07-05
Add cloudwatch query instructions to log in via the UX

Sgtpluck:
Added instructions for logging in via AWS directly, since that's easier if you know you need to switch roles.


zachmargolis (approved):
LGTM! Thanks for adding

PR #585, opened 2024-07-05
Remove Netlify support

zachmargolis:
**Why**: A recent supply-chain attack on polyfill-dot-io highlights the risks of linking to external domains for JS like we were doing here with unpkg.com

Ref: https://sansec.io/research/polyfill-supply-chain-attack

I do not think we are using Netlify for this site, so it seemed easier to remove the functionality entirely instead of an alternative approach like vendoring the JS


_articles/cloud-gov-pages.md
```diff
@@ -8,15 +8,14 @@ subcategory: Code
 
 ## Overview
 
-Login.gov uses [Cloud.gov Pages][cloud-gov-pages] as its hosting provider for our [static sites][static-sites]. By becoming a user of the Login.gov organization in Cloud.gov Pages, you'll have access to build logs and settings configuration for each of our sites. Organization access also allows you to log in on sites which use Netlify CMS ([including this handbook!][handbook-admin]).
+Login.gov uses [Cloud.gov Pages][cloud-gov-pages] as its hosting provider for our [static sites][static-sites]. By becoming a user of the Login.gov organization in Cloud.gov Pages, you'll have access to build logs and settings configuration for each of our sites.
```
aduth, line 11: I think the sentence is still valid for sites that _do_ continue to use Netlify CMS. We could just remove the snippet about the handbook admin.

```suggestion
Login.gov uses [Cloud.gov Pages][cloud-gov-pages] as its hosting provider for our [static sites][static-sites]. By becoming a user of the Login.gov organization in Cloud.gov Pages, you'll have access to build logs and settings configuration for each of our sites. Organization access also allows you to log in on sites which use Netlify CMS.
```

_articles/cloud-gov-pages.md
```diff
@@ -8,15 +8,14 @@ subcategory: Code
...
 
 [cloud-gov-pages]: https://cloud.gov/pages/
 [static-sites]: {% link _articles/github-gitlab.md %}#static-sites
-[handbook-admin]: {{ site.baseurl }}/admin/
 
 ## Requesting Access to Cloud.gov Pages
 
-**For those requesting access:** Reach out to any of the organization managers listed in the [Handbook Appendix][handbook-appendix-cloud-gov]. After you have been added to the organization, you will need to [sign in to the Cloud.gov Pages dashboard with GSA.gov][cloud-gov-pages-login] _before_ authenticating with Netlify CMS.
+**For those requesting access:** Reach out to any of the organization managers listed in the [Handbook Appendix][handbook-appendix-cloud-gov]. After you have been added to the organization, you will need to [sign in to the Cloud.gov Pages dashboard with GSA.gov][cloud-gov-pages-login].
```
aduth, line 18: I think I'd recommend either:

1. Keep the removed fragment because it's still true that someone would have to sign in to Cloud.gov Pages dashboard before accessing Netlify CMS on a site that uses it. (This article is not specific to the handbook)
2. Remove the second sentence altogether, since it's specifically referring to signing in for access to Netlify CMS. While that's true, this article is about Cloud.gov Pages broadly and not necessarily Netlify CMS. (Or at least there's not a clear distinction being made that you only have to do that if you intend to sign in to Netlify CMS)

```suggestion
**For those requesting access:** Reach out to any of the organization managers listed in the [Handbook Appendix][handbook-appendix-cloud-gov]. After you have been added to the organization, you will need to [sign in to the Cloud.gov Pages dashboard with GSA.gov][cloud-gov-pages-login] _before_ authenticating with Netlify CMS.
```

```suggestion
**For those requesting access:** Reach out to any of the organization managers listed in the [Handbook Appendix][handbook-appendix-cloud-gov].
```

aduth (approved):

PR #595, opened 2024-08-06
Add .nvmrc

matthinz:
Adds an .nvmrc file so that local node version management tools (nvm, asdf, etc.) can use the right version of node.

(I used the LTS version I currently have installed to pick a number.)


.nvmrc
```diff
@@ -0,0 +1 @@
+20.8.1
```
aduth, line 1: Could we use either `lts/*` to avoid having to update this, or at least let any version in the 20.x line work so that this doesn't fall out of date quickly and require someone to install an older, very specific version?

```suggestion
lts/*
```

```suggestion
20
```

References:

- https://github.com/18F/identity-idp/blob/main/.nvmrc
- https://github.com/GSA-TTS/identity-site/blob/main/.nvmrc
- https://github.com/18F/identity-design-system/blob/main/.nvmrc

aduth (approved):

.nvmrc
matthinz: TIL about `lts/*`!

PR #596, opened 2024-08-06
Add note about creating release in Gitlab for PKI

matthinz:
I think ultimately we'll want releases to sync to Github, but for now I guess we can just create in both places.

### DIFF
--- a/_articles/appdev-deploy.md
+++ b/_articles/appdev-deploy.md
@@ -257,7 +257,15 @@ Where `<PR_NUMBER>` is the number of the _merged_ PR.
    export GPG_TTY=$(tty)
    bin/tag-release
    ```
-2. Add release notes in GitHub:
+
+2. Add release notes in Gitlab:
+   1. Create a new release: <https://gitlab.login.gov/lg/identity-pki/-/releases/new>
+   2. Release title: `RC #{NUMBER}`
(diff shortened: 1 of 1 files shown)


aduth:
Is it necessary for us to create a release in GitLab?

matthinz:
Good question! I don't know. Probably not?

aduth:
@nick-soutouras Would you happen to know if it's possible / on the roadmap to sync GitLab releases to GitHub, so we could create releases in GitLab and let them sync over automatically?

There was [a related Slack discussion](https://gsa-tts.slack.com/archives/C0NGESUN5/p1718729481023609) not too far back, but it's unclear where we ended up with that.

matthinz:
I feel like this could probably be automated with a script. We probably want to create the _tag_ for the release in the Gitlab repo, but then create the "Release" in Github

nick-soutouras:
> @nick-soutouras Would you happen to know if it's possible / on the roadmap to sync GitLab releases to GitHub, so we could create releases in GitLab and let them sync over automatically?
> 
> There was [a related Slack discussion](https://gsa-tts.slack.com/archives/C0NGESUN5/p1718729481023609) not too far back, but it's unclear where we ended up with that.

I have an open issue to make this happen, will keep you in the loop (Ref issue #149 in Team Mary Gitlab)

PR #609, opened 2024-09-30
Add team Evelyn

zachmargolis:
Redo of @StaceyLEvans 's https://github.com/GSA-TTS/identity-handbook/pull/608, but from the main repo.

Also adds a commit to remove "product: true" because that comes with an Oncall Rotation requirement


kevinsmaster5 (approved):
Looks good!

PR #612, opened 2024-10-07
Document query-cloudwatch --sqlite option

matthinz:
Update the "Scripts" page with information about how to output event data from `query-cloudwatch` to a SQLite database.


aduth (approved):
Nice, this'll be a handy feature to have 👍

PR #619, opened 2024-11-04
Try to clarify production section of configuration, deployed environments and prod

mitchellhenke:
Based on discussion [here](https://gsa-tts.slack.com/archives/C03KDV9S8T1/p1720015058170999), there's quite a bit of ambiguity in environments and their interaction with different elements of configuration. This PR tries to disambiguate some, definitely open to suggestions.

### DIFF
--- a/_articles/appdev-feature-flags.md
+++ b/_articles/appdev-feature-flags.md
@@ -9,7 +9,7 @@ subcategory: Development
 
 ## Overview
 
-Our Rails applications use [configuration values][rails-configuration] to support per-environment configuration. This allows us to create new configuration values which control whether a specific application feature is enabled in an environment, so that we can incrementally iterate on a large feature over time before enabling it in lower environments and eventually the production environment. These are referred to as "feature flags", and are nothing more than boolean configuration values which are used in code to gate user and logic pathways.
+Our Rails applications use [configuration values][rails-configuration] to support per-environment configuration. This allows us to create new configuration values which control whether a specific application feature is enabled in an environment, so that we can iterate on a large feature over time while [rolling it out](#rolling-out-a-feature-flag). These are referred to as "feature flags", and are nothing more than boolean configuration values which are used in code to gate user and logic pathways.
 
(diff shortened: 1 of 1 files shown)


_articles/appdev-feature-flags.md
```diff
@@ -9,7 +9,7 @@ subcategory: Development
 
 ## Overview
 
-Our Rails applications use [configuration values][rails-configuration] to support per-environment configuration. This allows us to create new configuration values which control whether a specific application feature is enabled in an environment, so that we can incrementally iterate on a large feature over time before enabling it in sandbox and eventually production environments. These are referred to as "feature flags", and are nothing more than boolean configuration values which are used in code to gate user and logic pathways.
+Our Rails applications use [configuration values][rails-configuration] to support per-environment configuration. This allows us to create new configuration values which control whether a specific application feature is enabled in an environment, so that we can incrementally iterate on a large feature over time before enabling it in lower environments and eventually the production environment. These are referred to as "feature flags", and are nothing more than boolean configuration values which are used in code to gate user and logic pathways.
```
vrajmohan, line 12: should it be "before enabling it in higher environments"?

vrajmohan (commented):
Can we also say how `application.yml.default.prod` is used, as opposed to the `production` section of `application.yml.default`?

_articles/appdev-feature-flags.md
mitchellhenke: The process the paragraph is trying to get at is the order of:

1. enable for local development and testing
2. enable for lower (sandbox or staging) environments
3. enable for prod (secure.login.gov)

zachmargolis: what if we make that ordered list part of the docs?

mitchellhenke: That's discussed in the "Enabling a feature flag" section, so perhaps this part should be a bit more generic and link to that.

mitchellhenke: Made some bigger edits to re-organize that section if you all could take another look.

PR #620, opened 2024-11-04
Add new scenario to 50/50 docs

eileen-nava:
Inspired by this [github conversation](https://github.com/18F/identity-idp/pull/11440#discussion_r1825976279), I added another scenario to our 50/50 state documentation.

I haven't experienced this 50/50 state scenario with any of my previous PRs.  As a result, I'd appreciate input from anyone who has previously witnessed this scenario during a deploy.  Thanks! 🙏🏻

### DIFF
--- a/_articles/manage-50-50-state.md
+++ b/_articles/manage-50-50-state.md
@@ -139,6 +139,11 @@ of the cases above.
 * Deploy 1: Change the flag value
 * Make any changes to routes, jobs, and persistent data in following deploys as described above.
 
+## JavaScript Requests from API Endpoints
+We can hit 50/50 state errors when JavaScript requests to an API endpoint assume that the data will take a certain shape. If an older instance responds to the request, it could cause the request to fail.
+
+To avoid such an error, consider modifying the JavaScript request to be less coupled with the shape of the data that the API returns.
+
 ## Test in the 50/50 state
 When making a large change such as adding a new page to the Identity Verification flow, it can be
 helpful to test in the 50/50 state. Large changes are developed behind a feature flag and then
--- a/_articles/manage-50-50-state.md
+++ b/_articles/manage-50-50-state.md
@@ -142,7 +142,7 @@ of the cases above.
 ## JavaScript Requests from API Endpoints
 We can hit 50/50 state errors when JavaScript requests to an API endpoint assume that the data will take a certain shape. A newer or older instance may respond to the request in a way that the JavaScript isn't prepared to handle. This could cause unpredictable behavior, or may cause the code to throw an error if it assumes that a response property will exist.
 
-To avoid such an error, consider modifying the JavaScript to be less coupled with the shape of the data that the API returns. For example, you can define the response shape to define its properties as potentially `undefined`, which will also allow TypeScript type-checking to identify and potential errors that could be thrown.
+To avoid such an error, consider modifying the JavaScript to be less coupled with the shape of the data that the API returns. For example, you can define the response shape to define its properties as potentially `undefined`, which will also allow TypeScript type-checking to identify any potential errors that could be thrown.
 
 When introducing a new endpoint that you intend to request from, you can avoid 50/50 deployment issues by including the API endpoint in a separate, earlier deployment. Then you can safely assume that the endpoint will be available from your JavaScript code.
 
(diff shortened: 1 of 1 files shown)


WilliamBirdsall (approved):
Haven't run into this myself either but this doc change reads well :)

_articles/manage-50-50-state.md
```diff
@@ -139,6 +139,11 @@ of the cases above.
 * Deploy 1: Change the flag value
 * Make any changes to routes, jobs, and persistent data in following deploys as described above.
 
+## JavaScript Requests from API Endpoints
+We can hit 50/50 state errors when JavaScript requests to an API endpoint assume that the data will take a certain shape. If an older instance responds to the request, it could cause the request to fail.
```
shanechesnutt-ft, line 143: Not sure what the appropriate line length is here, but I think you need to split this line into multiple lines. This will be consistent with the rest of the file.

_articles/manage-50-50-state.md
```diff
@@ -139,6 +139,11 @@ of the cases above.
...
+
+To avoid such an error, consider modifying the JavaScript request to be less coupled with the shape of the data that the API returns.
```
shanechesnutt-ft, line 145: same thing here.

_articles/manage-50-50-state.md
aduth, line 143: It can also go the other direction, where a newer instance responds to the request, and the older JavaScript isn't prepared to handle that new response shape.

Maybe generalizing it a bit could help.

Also, I think the expected result isn't that the request would fail, but that either the behavior of the application is unpredictable, or the code might throw errors if it assumes a property exists that may not.

```suggestion
We can hit 50/50 state errors when JavaScript requests to an API endpoint assume that the data will take a certain shape. A newer or older instance may respond to the request in a way that the JavaScript isn't prepared to handle. This could cause unpredictable behavior, or may cause the code to throw an error if it assumes that a response property will exist.
```

_articles/manage-50-50-state.md
aduth, line 145: Maybe we can include an example of what it means to be decoupled.

Another option when adding new endpoints is to deploy the endpoint separately from the JavaScript, so that it's guaranteed to be there regardless of which version of the JavaScript is deployed.

```suggestion
To avoid such an error, consider modifying the JavaScript to be less coupled with the shape of the data that the API returns. For example, you can define the response shape to define its properties as potentially `undefined`, which will also allow TypeScript type-checking to identify and potential errors that could be thrown.

When introducing a new endpoint that you intend to request from, you can avoid 50/50 deployment issues by including the API endpoint in a separate, earlier deployment. Then you can safely assume that the endpoint will be available from your JavaScript code.
```

aduth (approved):
Thanks for writing this up! I think this'll be a very useful extra 50/50 case to look out for.

_articles/manage-50-50-state.md
eileen-nava: Thanks for the feedback!  Yeah, I agree with generalizing this.  Good point about the expected result, too.

_articles/manage-50-50-state.md
```diff
@@ -139,6 +139,13 @@ of the cases above.
 * Deploy 1: Change the flag value
 * Make any changes to routes, jobs, and persistent data in following deploys as described above.
 
+## JavaScript Requests from API Endpoints
+We can hit 50/50 state errors when JavaScript requests to an API endpoint assume that the data will take a certain shape. A newer or older instance may respond to the request in a way that the JavaScript isn't prepared to handle. This could cause unpredictable behavior, or may cause the code to throw an error if it assumes that a response property will exist.
+
+To avoid such an error, consider modifying the JavaScript to be less coupled with the shape of the data that the API returns. For example, you can define the response shape to define its properties as potentially `undefined`, which will also allow TypeScript type-checking to identify and potential errors that could be thrown.
```
aduth, line 145: Small typo in my earlier suggestion

```suggestion
To avoid such an error, consider modifying the JavaScript to be less coupled with the shape of the data that the API returns. For example, you can define the response shape to define its properties as potentially `undefined`, which will also allow TypeScript type-checking to identify any potential errors that could be thrown.
```

aduth (approved):

_articles/manage-50-50-state.md
eileen-nava: Thanks for the reminder to check line length.  I applied Andrew's suggestions, and I think line length is consistent now.

<details>
<summary>Screenshot with updated text:</summary>

50-50StateScenarioDocs

</details>

PR #588, opened 2024-07-24
Document the risks of running a migration without a recycle

eileen-nava:
Following up on [this conversation](https://gsa-tts.slack.com/archives/C0NGESUN5/p1721234296802419) in slack, I wanted to better document the risks of running a migration without a recycle.  Please let me know if there are additional risks that I did not document.  I mostly transcribed the information I learned via the linked slack exchange, which means this section may not be exhaustive.

PR #621, opened 2024-11-04
Document risks of no-migration recycles

eileen-nava:
This PR takes the place of #588.

### DIFF
--- a/_articles/appdev-deploy.md
+++ b/_articles/appdev-deploy.md
@@ -336,7 +336,7 @@ new configurations (config from S3).
     ```
 
 ### No-Migration Recycle
-When responding to a production incident with a config change, or otherwise in a hurry, you might want to recycle without waiting for a migration instance. For environments other than prod, note that if a migration has been introduced on `main`, new instances will fail to start until migrations are run.
+When responding to a production incident with a config change, or otherwise in a hurry, you might want to recycle without waiting for a migration instance.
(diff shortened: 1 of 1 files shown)


WilliamBirdsall (approved):

_articles/appdev-deploy.md
```diff
@@ -346,3 +346,10 @@ aws-vault exec prod-power -- ./bin/asg-recycle prod idp --skip-migration
 ```bash
 aws-vault exec prod-power -- ./bin/scale-remove-old-instances prod ALL
 ```
+
+#### Risks with No-Migrations Recycle
+For environments other than prod, note that if a migration has been introduced on `main`, new instances will fail to start until migrations are run.
+
+Additionally, migration instances are responsible for compiling assets.  If assets have changed since the last migration, we recommend against running a no-migration recycle.  Otherwise, the new servers will look for fingerprinted asset files that don't exist, because the migration instances never created them.
+
+To tell if assets have changed since the last migration, inspect the [Environments status page](https://dashboard.int.identitysandbox.gov/env).  Click "pending changes" and determine if any files changed in `app/assets/stylesheets` or `app/javascript`.
```
gina-yamada, line 355: Thanks for adding the last sentence in. I think it is helpful information.

gina-yamada (approved):

PR #635, opened 2024-12-10
Use Pagefind for search

aduth:
## 🛠 Summary of changes

Updates to use [Pagefind](https://pagefind.app/) as a search tool in place of `simple-jekyll-search`, since it works well out of the box to provide contextual information for search results.

## 👀 Screenshots

Default|With search results
---|---
image|image


zachmargolis (approved):
this works surprisingly well! should we ditch the other jekyll search plugin we have for the top right of the navbar?

aduth:
>this works surprisingly well!

Yeah it was a very pleasant quickstart experience and required surprisingly little code. Really the only pain point here is that the NPM `pages` script is run _before_ Jekyll, and since the tool needs the site to be built to index, that's the reason for the Jekyll plugin.

>should we ditch the other jekyll search plugin we have for the top right of the navbar?

Yeah, I think we would. I'd want to think about how best to enable searching while already within an article. I'm personally okay with the default styling for the homepage, but maybe we can find a way to style it more compactly like what we have in the top-right currently.

aduth:
Rev2:

Default|With search
---|---
image|image

aduth:
Marking this ready for review! It came together pretty nicely I think.

vrajmohan:
Looks great. I'm testing it locally.

vrajmohan (approved):
Works like a charm. Later, we can look into influencing the search results...

aduth:
>Later, we can look into influencing the search results...

Yeah, there's some options to tinker with, both in appearance and weighting. The [`showSubResults` option](https://pagefind.app/docs/sub-results/#showing-sub-results-with-the-default-ui), for example, adds a lot of detail, but IMO it's a bit _too much_ detail for each entry? The default options also show an image, but most of our pages don't really have a useful image to use at the moment.

image

aduth:
Also, these changes should be replicated over to the internal handbook as well. I might make that a Friday task.

PR #471, opened 2023-09-27
LG-10115 - Document IdV rate limiting

jmax-gsa:

PR #418, opened 2023-04-20
Pauldoom/ir checklist reboot

pauldoomgov:
This is part 1 of a N part refinement of the checklist and IR guide.

The main purpose here: Restructure the checklist to allow each role holder to focus on their responsibilities.

PR #649, opened 2024-12-12
Remove unused images

aduth:
## 🛠 Summary of changes

Removes unused images from source code.

Discovered with a similar script as in https://github.com/18F/identity-idp/pull/11215:

```bash
for i in `ls images`
do
  if ! grep -q -R _articles -e $i
  then
    echo $i;
  fi
done
```

declare-incident-workflow-1.png: Added in #418, never used
declare-incident-workflow-2.png: Added in #418, never used
declare-incident-workflow-3.png: Added in #418, never used
idv-session-errors-failure.png: Added in #471, never used
idv-ssn-rate-limited.png: Added in #471, never used
statuspage-sample-message.png: Last reference removed in #389

## 📜 Testing Plan

Verify removed images aren't referenced anywhere.


vrajmohan (approved):
Looks good

PR #651, opened 2024-12-12
Improve searchability of redirect pages

aduth:
## 🛠 Summary of changes

Updates the new search behavior to improve searchability of redirect links.

We have a number of external links as stub articles, and these are currently hard to search for, e.g. "org chart".

Changes include:

- Extend [`jekyll-redirect-from` base redirect template](https://github.com/jekyll/jekyll-redirect-from/blob/master/lib/jekyll-redirect-from/redirect.html) to include annotations to enable search indexing
- And (and require) descriptions for redirect page stubs
- Exclude category pages from indexing

## 📜 Testing Plan

1. Go to preview URL
2. Search for a redirect link, e.g. "org chart"
3. See useful results

## 👀 Screenshots

Before|After
---|---
image|image


_articles/piv-ssh.md
aduth, line 1: Removed because I was going to need to come up with a `description`, but discovered that the content is outdated and no longer valid.

Related Slack discussion: https://gsa-tts.slack.com/archives/C16RSBG49/p1734039506543219

_layouts/redirect.html
```diff
@@ -0,0 +1,24 @@
+---
+---
+<!DOCTYPE html>
```
zachmargolis, line 3: should we add a source comment that this overrides jekyll-redirect_from?
```suggestion
---
---
{% comment %}
  This overrides the template that jekyll_redirect-from uses so that we can make descriptions searchable in the site index
{% endcomment %}
<!DOCTYPE html>
```

spec/markdown_spec.rb
```diff
@@ -33,6 +33,8 @@ def category_exists?(title)
           expect(frontmatter).to be
 
           aggregate_failures do
+            expect(frontmatter['description']).to be
```
zachmargolis, line 36: empty string would still pass this, right? like I could be a maliciously compliant page author with:

```
description: ""
```

```suggestion
            expect(frontmatter['description']).to_not be_empty
```

zachmargolis (approved):

_layouts/redirect.html
aduth: Yep, we should 👍 Added in 7355547

PR #652, opened 2024-12-12
Flag and suggest alternatives for common mispellings

aduth:
## 🛠 Summary of changes

Updates spell checker to consider common misspellings of words, including a suggested alternative.

**Draft:**

- [ ] Fix existing issues
- [ ] Any other to flag?

## 📜 Testing Plan

Verify build passes, and that revised content changes look sensible.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -57,6 +57,14 @@
     "FICAM",
     "FPKI"
   ],
+  "flagWords": [
+    "Github->GitHub",
+    "Gitlab->GitLab",
+    "pki->PKI",
+    "Idp->IdP",
+    "idp->IdP",
+    "IDP->IdP"
+  ],
   "patterns": [
     {
       "name": "Markdown links",
--- a/cspell.json
+++ b/cspell.json
@@ -126,6 +126,10 @@
       "name": "Liquid Syntax With Dashes",
       "pattern": "\\{[%]-? (.*) -?[%]\\}",
       "description": "Some of the GitHub docs have this syntax"
+    },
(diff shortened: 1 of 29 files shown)


cspell.json
```diff
@@ -57,6 +57,14 @@
     "FICAM",
     "FPKI"
   ],
+  "flagWords": [
+    "Github->GitHub",
+    "Gitlab->GitLab",
+    "pki->PKI",
+    "Idp->IdP",
+    "idp->IdP",
+    "IDP->IdP"
```
mitchellhenke, line 66:
```suggestion
    "IDP->IdP",
    "Login->Login.gov"
```

mitchellhenke (approved):

cspell.json
zachmargolis: Will `"login.gov->Login.gov"` flag URLs?

zachmargolis, line 66: what about making all-caps "IDP" the preferred spelling?

mitchellhenke: Good question, it looks like punctuation is ignored, so it can't detect `login.gov` anyway

mitchellhenke: https://cspell.org/docs/how-it-works/#things-to-note

aduth: Yeah, unfortunately "Login" catches correctly-spelled "Login.gov" as well because of the dot.

zachmargolis: dang! That is a bummer, but I guess it makes sense to match the industry

PR #653, opened 2024-12-20
Add test coverage for ViewComponent plugin

aduth:
## 🛠 Summary of changes

Adds some basic test coverage for the Jekyll + view_component plugin added in #628 and iterated in subsequent pull requests.

## 📜 Testing Plan

```
rspec spec/plugins/view_component_spec.rb 
```

### DIFF
--- a/spec/plugins/view_component_spec.rb
+++ b/spec/plugins/view_component_spec.rb
@@ -1,0 +1,125 @@
+require 'jekyll'
+require 'liquid'
+require_relative '../../_plugins/view_component.rb'
+
+RSpec.describe ComponentTag do
+  subject(:rendered) do
+    converter.convert(Liquid::Template.parse(content).render!(payload, { registers: { site: } }))
+  end
+  let(:payload) { {} }
+  let(:content) { '' }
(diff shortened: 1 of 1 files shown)


spec/plugins/view_component_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+  context 'with string param' do
+    let(:content) { '{% component test arg="string" %}content{% endcomponent %}' }
+    let(:test_component) do
+      Class.new(BaseComponent) do
+        def initialize(arg:)
+          @arg = arg
+        end
+
+        def call
+          tag.div(@arg.class.name + @arg.to_s)
```
zachmargolis, line 58: WDYT of making the test content like this?
```suggestion
          tag.div("arg_class=#{@arg.class.name} arg=#{@arg}")
```
then the string to match would be like:
```
<div>arg_class=String arg=string</div>
```
might be a bit easier to read?

zachmargolis (approved):

spec/plugins/view_component_spec.rb
aduth: >WDYT of making the test content like this?

Sounds good to me! Updated in 33b67a6.

PR #666, opened 2025-02-20
Use jekyll link tag for internal links instead of hardcoding production URLs

mitchellhenke:
## 🛠 Summary of changes

Makes a couple small fixes to prefer usage of `link` with a page rather than `handbook.login.gov`.

### DIFF
--- a/_articles/appdev-oncall-guide.md
+++ b/_articles/appdev-oncall-guide.md
@@ -108,7 +108,7 @@ Refer to [Troubleshooting expiring PIV/CAC certs]({% link _articles/troubleshoot
 
 ## Response Times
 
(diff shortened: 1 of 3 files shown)


aduth (approved):
LGTM :+1:

I wonder if we could have an automated test to check for hardcoded links?

mitchellhenke:
> LGTM 👍
> 
> I wonder if we could have an automated test to check for hardcoded links?

Yeah, I think so. There are some more documentation-y types of links to the handbook home page that we may not want to use `link` for. It could be a config/variable instead so we can check for hardcoded "handbook.login.gov" more easily.

PR #731, opened 2026-09-22
Add details to pwned password maintenance instructions

zombonic-tts:
When doing Team Katherine oncall maintenance task of updating our pwned password list, I noticed a couple helpful details that can be added to our documentation.


joshuaalley25 (approved):
Thanks for adding to this.

